PR #161, opened 2021-10-07
Deduplicate type name

### LINKED ISSUE #160: Awkward API names for input args [Python, Go, C#]
## Hello!

## Issue details

Hi, this may be a duplicate but I find this really annoying especially in python.

The naming scheme for the `args`/`props` of the resources is really awkward in Python, Go or C#. The `*Args` objects tend to get extremely long names which in python often breaks maximum line length (default is 79 chars). Therefore PEP20 (`zen of python` ) suggests using namespaces. I believe this is probably true of other affected languages as well.

`*Args` classes contain duplication in the names just to prevent a clash in a single namespace.

Example:

TS:
```typescript
import * as aws_native from "@pulumi/aws-native";

const cloudfrontdistribution = new aws_native.cloudfront.Distribution("cloudfrontdistribution", {
    distributionConfig: {
        cacheBehaviors: [{
```

Python: 
Notice `DistributionDistribution`.
```python
cloudfrontdistribution = aws_native.cloudfront.Distribution("cloudfrontdistribution",
    distribution_config=aws_native.cloudfront.DistributionDistributionConfigArgs(
        cache_behaviors=[aws_native.cloudfront.DistributionCacheBehaviorArgs(
```

In such cases in python, this could be solved by creating a package to break the namespace.
Seeing that there are 2 types of distribution configs: `'DistributionDistributionConfigArgs'` and `StreamingDistributionTagArgs`.
I noticed that there are already 2 modules `distribution.py` and `streaming_distribution.py` so I do not understand the need for `_inputs.py` that requires this awkward naming schema, while the respective `*Args` classes could live in their respective files.

Go:
Notice `DistributionDistribution`.
```go
pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := cloudfront.NewDistribution(ctx, "cloudfrontdistribution", &cloudfront.DistributionArgs{
			DistributionConfig: &cloudfront.DistributionDistributionConfigArgs{
				CacheBehaviors: cloudfront.DistributionCacheBehaviorArray{
```

C#:
```c#
class MyStack : Stack
{
    public MyStack()
    {
        var cloudfrontdistribution = new AwsNative.CloudFront.Distribution("cloudfrontdistribution", new AwsNative.CloudFront.DistributionArgs
        {
            DistributionConfig = new AwsNative.CloudFront.Inputs.DistributionDistributionConfigArgs
            {
                CacheBehaviors = 
                {
                    new AwsNative.CloudFront.Inputs.DistributionCacheBehaviorArgs
```

### Affected area/feature
* Codegen

mikhailshilkov:
Do not prefix a type name with the resource name if the resource name is already a prefix of the CF type name. Fix #160


lblackstone (approved):
🎉 

Some of those generated names were pretty funny. Thanks for cleaning this up!

PR #191, opened 2021-10-29
Add additional keywords to the Pulumi schema spec

praneetloke:
Similar to https://github.com/pulumi/pulumi-azure-native/pull/1254 and https://github.com/pulumi/pulumi-google-native/pull/219.

Updated `github.com/pulumi/pulumi/v3/pkg` and `...sdk` and ran the codegen steps. I separated the commits based on folder paths. I do see that the submodule reference for ` aws-cloudformation-user-guide
Public` shows up as having new commits. I assume we need to update the submodule reference? Let me know if I should updated that to the latest ref?

Screen Shot 2021-10-29 at 18 24 44


mikhailshilkov:
I'd prefer if you didn't update the JSON specs in the same PR as updating p/p as this makes review difficult.

> I assume we need to update the submodule reference?

I'd say we should _not_ update the submodule here. When you clone the repo, you should run `make ensure` but don't run `make discovery` or `make update_submodules`. Happy to do this with you if that's easier.

praneetloke:
Closing as Levi is going to have a new PR to replace this.

> I'd say we should not update the submodule here. When you clone the repo, you should run `make ensure` but don't run `make discovery` or `make update_submodules`. Happy to do this with you if that's easier.

Got it! Thanks for the info. What's the proper way to regenerate the `schema.json` file since I would have wanted the change to the schema spec in Go to be reflected in the serialized schema file?

mikhailshilkov:
> What's the proper way to regenerate the schema.json file since I would have wanted the change to the schema spec in Go to be reflected in the serialized schema file?

```
make codegen generate_schema
```

PR #237, opened 2021-11-20
Avoid potentially spilling credentials to state

### LINKED ISSUE #236: Custom providers can leak credentials to state file from environment variables
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

If using `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` as environment variables and using a custom (non-default) provider, the contents of these environment variables can be leaked into the state file in clear text.

### Steps to reproduce

1. Start with a program similar to this:
```
import * as awsnative from "@pulumi/aws-native";

const prov = new awsnative.Provider("prov", {region: "us-west-2"});
// Create an AWS resource (S3 Bucket)
const bucket1 = new awsnative.s3.Bucket("my-bucket1", {}, {provider: prov});
const bucket2 = new awsnative.s3.Bucket("my-bucket2", {});

// Export the name of the bucket
export const bucket1Name = bucket1.id;
export const bucket2Name = bucket2.id;
```

2. Export the above environment variables with AWS credentials. `export AWS_SECRET_ACCESS_KEY=xxyyzz`
3. Run a `pulumi up` and dump the resulting state - `pulumi stack export | grep accessKey`
4. Notice `xxyyzz` in plain text.

viveklak:
Fixes #236

### DIFF
--- a/provider/pkg/schema/gen.go
+++ b/provider/pkg/schema/gen.go
@@ -135,6 +135,94 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 		Provider: pschema.ResourceSpec{
 			ObjectTypeSpec: pschema.ObjectTypeSpec{
 				Description: "The provider type for the AWS native package. By default, resources use package-wide configuration settings, however an explicit `Provider` instance may be created and passed during resource construction to achieve fine-grained programmatic control over provider settings. See the [documentation](https://www.pulumi.com/docs/reference/programming-model/#providers) for more information.",
+				Properties: map[string]pschema.PropertySpec{
+					"accessKey": {
+						Description: "The access key for API operations. You can retrieve this from the ‘Security & Credentials’ section of the AWS console.",
+						TypeSpec:    pschema.TypeSpec{Type: "string"},
+						Secret:      true,
+					},
+					"allowedAccountIds": {
+						Description: "List of allowed AWS account IDs to prevent you from mistakenly using an incorrect one. Conflicts with `forbiddenAccountIds`.",
+						TypeSpec:    pschema.TypeSpec{Type: "array", Items: &pschema.TypeSpec{Type: "string"}},
+					},
+					"assumeRole": {
+						Description: "Configuration for retrieving temporary credentials from the STS service.",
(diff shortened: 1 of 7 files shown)


provider/pkg/schema/gen.go
```diff
@@ -132,104 +135,9 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 		Provider: pschema.ResourceSpec{
 			ObjectTypeSpec: pschema.ObjectTypeSpec{
 				Description: "The provider type for the AWS native package. By default, resources use package-wide configuration settings, however an explicit `Provider` instance may be created and passed during resource construction to achieve fine-grained programmatic control over provider settings. See the [documentation](https://www.pulumi.com/docs/reference/programming-model/#providers) for more information.",
-				Properties: map[string]pschema.PropertySpec{
```
viveklak, line 135: I believe the Properties defined here are redundant given the input properties below.

mikhailshilkov: `Properties` are outputs of the provider resource. Don't we define those in other provides?

viveklak: Doesn't seem like it? See https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/gen/schema.go#L95 and https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/gen/schema.go#L95.

mikhailshilkov: AWS does have them:
image

viveklak: I removed the properties change for now. I did mark the 3 sensitive fields as secrets though.

viveklak:
Adding @pgavlin to make sure we are being consistent with the spirit of https://github.com/pulumi/pulumi-azure/pull/583/ which was the last incarnation of this issue.

mikhailshilkov (approved):
I would be okay to exclude the secret properties from provider outputs - I don't think anyone would expect to get them as outputs

pgavlin (approved):

viveklak:
> I would be okay to exclude the secret properties from provider outputs - I don't think anyone would expect to get them as outputs

Dropped the secret fields.

PR #346, opened 2022-02-04
Implement get functions for all resources types

### LINKED ISSUE #83: Implement get functions for all resources
Similarly to other providers, there should be a `getFoo` function for each `Foo` resource

danielrbradley:
Fixes #83 

Known limitations:
- Ignores resources where the identifier is a referenced sub-type (only `S3::StorageLens`)
- Excludes `MediaConnect::FlowOutput` as `getFlowOutput` clashes with the existing `MediaConnect::Flow` method of the same name (see: https://github.com/pulumi/pulumi/issues/8886)


viveklak:
@danielrbradley thanks for the PR! Could you rebase the PR with a commit for the codegen changes and one each for the SDKs in each language? See azure-native's PRs for a reference. GH UI is unworkable at a certain scale otherwise.

examples/get-ts/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws-native";
+
+const logGroup = new aws.logs.LogGroup("log-test", {
+  retentionInDays: 90,
+});
+
+const fetchedLogGroup = aws.logs.getLogGroupOutput({
+  logGroupName: logGroup.logGroupName.apply((n) => n ?? ""),
```
viveklak, line 11: Since you are using the fnOutput variant of the getter, do you need the apply here?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## HEAD (Unreleased)
 
+- Implement get functions for all resources
```
viveklak, line 3: Consider adding a link to this PR here.

provider/pkg/provider/provider.go
```diff
@@ -453,6 +459,35 @@ func (p *cfnProvider) Invoke(ctx context.Context,
...
+	cf, ok := p.resourceMap.Functions[tok]
+	if !ok {
+		return nil, false
+	}
+	return func(p *cfnProvider, ctx context.Context, inputs resource.PropertyMap) (resource.PropertyMap, error) {
+		idParts := make([]string, len(cf.Identifiers))
+		for i, v := range cf.Identifiers {
+			pv, ok := inputs[resource.PropertyKey(v)]
+			if !ok {
+				return nil, errors.Errorf("missing identifier property '%s'", v)
```
viveklak, line 472: Nit: use %q instead of `'%s'`

provider/pkg/provider/provider.go
```diff
@@ -453,6 +459,35 @@ func (p *cfnProvider) Invoke(ctx context.Context,
...
+			}
+			if !pv.IsString() {
+				return nil, errors.Errorf("identifier property '%s', expected type string, found %s", v, pv.TypeString())
```
viveklak, line 475: `%q`

provider/pkg/provider/provider.go
```diff
@@ -453,6 +459,35 @@ func (p *cfnProvider) Invoke(ctx context.Context,
...
+			}
+			idParts[i] = pv.StringValue()
+		}
+		identifier := strings.Join(idParts, "|")
+		fmt.Printf("Invoking %s", cf.CfType)
```
viveklak, line 480: You probably want this in a verbose logging option instead?

provider/pkg/provider/provider.go
```diff
@@ -453,6 +459,35 @@ func (p *cfnProvider) Invoke(ctx context.Context,
...
+			pv, ok := inputs[resource.PropertyKey(v)]
+			if !ok {
+				return nil, errors.Errorf("missing identifier property '%s'", v)
+			}
+			if !pv.IsString() {
+				return nil, errors.Errorf("identifier property '%s', expected type string, found %s", v, pv.TypeString())
+			}
+			idParts[i] = pv.StringValue()
+		}
+		identifier := strings.Join(idParts, "|")
```
viveklak, line 479: Yuck - I didn't realize this is how cctl handles composite identifiers...

viveklak (changes requested):
Generally looks good. Some minor changes requested above.

examples/get-ts/index.ts
danielrbradley: It's because it can be null - and in Typescript that's a compilation failure -> `Output<string | undefined> cannot be assigned to Input<string>`

provider/pkg/provider/provider.go
danielrbradley: Yeah, kinda nasty, but 🤷

viveklak (approved):
LGTM

PR #456, opened 2022-04-18
[provider] Implement getUrlSuffix.

pgavlin:
Keep a mapping from regions to partitions and partitions to URL
suffixes and discover the partition for the configured region during
startup time. getUrlSuffix then returns the URL suffix for the
configured partition.

### DIFF
--- a/provider/pkg/provider/partitions.go
+++ b/provider/pkg/provider/partitions.go
@@ -1,0 +1,62 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


provider/pkg/provider/partitions.go
```diff
@@ -0,0 +1,62 @@
...
+		RegionRegex: regexp.MustCompile("^us\\-iso\\-\\w+\\-\\d+$"),
+	},
+	{
+		ID:          "aws-iso-b",
+		URLSuffix:   "sc2s.sgov.gov",
+		RegionRegex: regexp.MustCompile("^us\\-isob\\-\\w+\\-\\d+$"),
+	},
+}
+
+func getPartition(region string) partition {
```
viveklak, line 55: Nit - add some quick unit tests while we are at it?

PR #493, opened 2022-05-16
Add support for passing AWS IAM Role to Cloud Control API

alex-berger:
AWS Cloud Control API (and thus AWS CloudFormation) allows passing an AWS IAM Role to all resource modification actions (`CreateResource`, `UpdateResource`, `DeleteResource`, `GetResource`). The passed IAM Role is then used by the Cloud Control API to access, create, update or delete the AWS resources. With this PR an (optional) AWS IAM Role can be defined in the Provider configuration, which will then be passed to Cloud Control API.

Note, this is a unique feature for server side security enforcement, not to be confused with the already existing support for [`assumeRole`](https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/provider/provider.go#L348), which is used to obtain temporary client credentials.

### DIFF
--- a/provider/pkg/schema/gen.go
+++ b/provider/pkg/schema/gen.go
@@ -50,6 +50,10 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 					Description: "Configuration for retrieving temporary credentials from the STS service.",
 					TypeSpec:    pschema.TypeSpec{Ref: "#/types/aws-native:config:AssumeRole"},
 				},
+				"roleArn": {
+					Description: "The Amazon Resource Name (ARN) of the AWS Identity and Access Management (IAM) role for Cloud Control API to use when performing this resource operation. Note, this is a unique feature for server side security enforcement, not to be confused with assumeRole, which is used to obtain temporary client credentials. If you do not specify a role, Cloud Control API uses a temporary session created using your AWS user credentials instead.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
 				"defaultTags": {
 					Description: "Configuration block with resource tag settings to apply across all resources handled by this provider. This is designed to replace redundant per-resource `tags` configurations. Provider tags can be overridden with new values, but not excluded from specific resources. To override provider tag values, use the `tags` argument within a resource to configure new tag values for matching keys.",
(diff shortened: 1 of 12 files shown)


provider/pkg/schema/gen.go
```diff
@@ -50,6 +50,10 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 					Description: "Configuration for retrieving temporary credentials from the STS service.",
 					TypeSpec:    pschema.TypeSpec{Ref: "#/types/aws-native:config:AssumeRole"},
 				},
+				"role": {
+					Description: "The Amazon Resource Name (ARN) of the AWS Identity and Access Management (IAM) role for Cloud Control API to use when performing this resource operation.",
```
viveklak, line 54:
```suggestion
					Description: "The Amazon Resource Name (ARN) of the AWS Identity and Access Management (IAM) role for Cloud Control API to use when performing this resource operation. Note, this is a unique feature for server side security enforcement, not to be confused with assumeRole, which is used to obtain temporary client credentials. If you do not specify a role, Cloud Control API uses a temporary session created using your AWS user credentials instead.",
```

provider/pkg/schema/gen.go
```diff
@@ -144,6 +148,10 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 						Description: "Configuration for retrieving temporary credentials from the STS service.",
 						TypeSpec:    pschema.TypeSpec{Ref: "#/types/aws-native:index:ProviderAssumeRole"},
 					},
+					"role": {
```
viveklak, line 151: Nit: switch to `roleArn`?

provider/pkg/schema/gen.go
```diff
@@ -144,6 +148,10 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
...
+						Description: "The Amazon Resource Name (ARN) of the AWS Identity and Access Management (IAM) role for Cloud Control API to use when performing this resource operation.",
```
viveklak, line 152: See description suggestion above.

viveklak (commented):
Some minor suggestions. This looks great otherwise and thanks for making the PR! Would you be able to rebase the changelog as well please?

viveklak (approved):

PR #598, opened 2022-08-23
[internal] Adding Java builds to the SDK

stack72:


t0yv0 (approved):
Java compilation looking better!

PR #678, opened 2022-11-02
Fix updates to write-only properties

### LINKED ISSUE #277: Lambda Function update fails
Given the following resource deployed

```ts
const ping = new awsnative.lambda.Function("ping", {
  role: role.arn,
  runtime: "nodejs14.x",
  handler: "index.handler",
  code: {
      zipFile: `
      exports.handler = function(event, context, callback){
        callback(null, "TODO");
      };`,
    },
  functionName: "ping-name2",
});
```

I then try changing the `handler` property and update the resource. The update fails though:

> CloudControl: UpdateResource, https response error StatusCode: 400, RequestID: 6912f4f9-211b-4620-9834-0670b25ae58e, api error ValidationException: Model validation failed (#: required key [Code] not found)

Our patch package looks like this:

```
{"ClientToken":"3e99641d-2608-4ef2-9640-f1d8bfe3140e","Identifier":"ping-name2",
"PatchDocument":"[{\"op\":\"replace\",\"path\":\"/Handler\",\"value\":\"index.handler2\"}]","TypeName":"AWS::Lambda::Function"}
```

which looks reasonable to me.

I tried changing `code` as well, but then I get a different error:

> CloudControl: UpdateResource, https response error StatusCode: 400, RequestID: 76597d5c-93b5-472b-b48f-e6990eadb2fc, api error ValidationException: [REPLACE Operation] noSuchPath in source, path provided : //Code

Our patch has `/Code` not `//Code`:

```
{"ClientToken":"cd2e711f-02e6-4f0c-8fc8-fe867abbbd4b","Identifier":"ping-name2",
"PatchDocument":"[{\"op\":\"replace\",\"path\":\"/Code\",\"value\":{\"ZipFile\":\"\\n      exports.handler = function(event, context, callback){\\n        callback(null, \\\"TODO2\\\");\\n      };\"}}]","TypeName":"AWS::Lambda::Function"}
```

Is that an upstream issue or is the provider doing something wrong?

danielrbradley:
Check if properties are write-only, and use an "add" if so. This is because the old values of write only properties don't show up when applying the diff within CC, so we need to do an "add".

Fixes #277 

This affects ~370 properties across 123 resources

### DIFF
--- a/examples/update/step1/index.ts
+++ b/examples/update/step1/index.ts
@@ -1,5 +1,6 @@
 // Copyright 2016-2021, Pulumi Corporation.
-
+import * as pulumi from "@pulumi/pulumi";
+import * as awsClassic from "@pulumi/aws";
 import * as aws from "@pulumi/aws-native";
 
 const logGroup = new aws.s3.Bucket("bucket", {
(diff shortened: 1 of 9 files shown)


examples/update/step2/index.ts
```diff
@@ -8,3 +9,35 @@ const logGroup = new aws.s3.Bucket("bucket", {
...
+                    "logs:PutLogEvents",
+                ],
+                Resource: "arn:aws:logs:*:*:*",
+            }],
+        })
+    }]
+});
+
+const func = new aws.lambda.Function("function", {
+    runtime: "nodejs12.x",
```
jkodroff, line 32: Might want to use a more recent runtime so we don't have to fix unexpectedly if they drop support. This is the oldest of the 3 supported node runtimes - 16.x is most recent.

provider/pkg/schema/convert.go
```diff
@@ -135,6 +142,15 @@ func (c *sdkToCfnConverter) valueToPatch(opName, propName string, prop pschema.P
 	return op
 }
 
+func (c *sdkToCfnConverter) isPropWriteOnly(cfnName string) bool {
```
jkodroff, line 145: Tests for this, maybe, if easy? Arguably not worth it. IDK.

provider/pkg/schema/convert.go
```diff
@@ -135,6 +142,15 @@ func (c *sdkToCfnConverter) valueToPatch(opName, propName string, prop pschema.P
...
+	for _, createOnlyProp := range c.spec.WriteOnly {
```
thomas11, line 146: I assume there are not that many properties to make using a map worthwhile?

thomas11 (approved):

examples/update/step1/index.ts
```diff
@@ -8,3 +9,47 @@ const logGroup = new aws.s3.Bucket("bucket", {
...
+    handler: "index.handler",
+    code: {
+        zipFile: `
+exports.handler =  async function(event, context) {
+  console.log("Running")
+  return "foo"
+}
+`,
+    },
+}, { dependsOn: new DelayResource("delay", { ms: 60 * 1000 }, { dependsOn: lambdaRole }) });
```
thomas11, line 55: Is there no way to actually check whether the function is ready?

thomas11 (approved):

PR #713, opened 2022-11-29
Manually implement SnapStart

danielrbradley:
Tentative change in case the specs don't get released within a reasonable timeframe.

This would immediately get overwritten the next time discovery runs, so a risk here is that it blocks us updating the provider until they ship the spec. Perhaps we should seek input on how long this might be.

The second risk is that the spec might vary in some subtle way which could cause breaking changes for early adopters. I would consider this a lesser risk.


aws-cloudformation-schema/aws-lambda-function.json
```diff
@@ -24,6 +24,17 @@
   "additionalProperties" : false,
   "primaryIdentifier" : [ "/properties/FunctionName" ],
   "definitions" : {
+    "SnapStartConfig": {
```
thomas11, line 27: What did you base this addition on, when the CF schema isn't released yet?

danielrbradley: A bit of guessing work based on advanced documentation releases plus Josh tested it against the early access we had enabled on our account and it seemed to work fine.

thomas11: Thanks. Just realized I can't approve it anyway since it's a draft.

danielrbradley:
This is now available in the main schema

PR #726, opened 2022-12-06
Make supported resources append-only

danielrbradley:
Don't automatically remove types once they've been published.

### DIFF
--- a/provider/cmd/pulumi-gen-aws-native/main.go
+++ b/provider/cmd/pulumi-gen-aws-native/main.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/pulumi/pulumi/pkg/v3/codegen"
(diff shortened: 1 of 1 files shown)


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -157,6 +158,9 @@ func readSupportedResourceTypes(outDir string) []string {
 }
 
 func writeSupportedResourceTypes(outDir string) error {
+	supported := readSupportedResourceTypes(outDir)
```
thomas11, line 161: I think this needs a comment on why we're reading the previous ones and merge.

danielrbradley:
@thomas11 I've added a short explanation of the process in the comments and am now writing any types we're still maintaining but have been removed by AWS into `deprecated-types.txt` so it's visible in the nightly PR.

PR #746, opened 2022-12-19
Implement missing getPartition function

### LINKED ISSUE #742: Documented `getPartition` function not found error
### What happened?

According to the API documentation, the aws-native package has the `getPartition` function (see https://www.pulumi.com/registry/packages/aws-native/api-docs/getpartition/), just like the classic aws package. However, when you call the function you receive an error that the function is unknown.

### Steps to reproduce

```

const partition = aws.getPartition();
```

### Expected Behavior

The actual partition to be returned, like within the aws-classic package

### Actual Behavior

``` Error: invocation of aws-native:index:getPartition returned an error: unknown function 'aws-native:index:getPartition'```

### Output of `pulumi about`

```
Dependencies:
NAME                VERSION
@pulumi/aws-native  0.44.0
@pulumi/pulumi      3.49.0
```

### Additional context

_No response_

### Contributing

danielrbradley:
This was previously generated in the schema but the "pseudo parameters" have to be implemented manually, which partition was not.

- Add the partition DNS Suffix to the output as it's a crucial piece of information for users.
- Implement the partition string property as the identifier.
- Include basic documentation on what the properties will look like.

Matching the design of the classic provider - using DNS instead of URL. The source provider object, and therefore URL, is only an internal construct and not a model derived from AWS directly, so DNS seems to be a better choice for the external interface.

Fixes #742

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -477,6 +477,7 @@ var functions = map[string]func(*cfnProvider, context.Context, resource.Property
 	"aws-native:index:getSsmParameterString": (*cfnProvider).getSSMParameterString,
 	"aws-native:index:getSsmParameterList":   (*cfnProvider).getSSMParameterList,
 	"aws-native:index:importValue":           (*cfnProvider).importValue,
+	"aws-native:index:getPartition":          (*cfnProvider).getPartition,
 }
 
 type invokeFunc func(p *cfnProvider, ctx context.Context, inputs resource.PropertyMap) (resource.PropertyMap, error)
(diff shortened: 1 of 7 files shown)


mikhailshilkov:
Looks great! Should we add this invoke call to a test?

danielrbradley:
@mikhailshilkov yeah ideally we'd have all of these tested but program test doesn't feel like the right method for this. If we had a lightweight unit tests setup then this would be much easier. I ran this as a local test having reproduced the original user issue and confirmed this solved it.

roothorp (approved):

PR #807, opened 2023-02-14
Add naming trivia implementation to autonaming process

### LINKED ISSUE #794: FIFO Queue does not properly create a name ending with `.fifo`
### What happened?

I deployed a FIFO Queue that creates properly in CFN but fails complaining about an invalid name. I assume something is broken with how a name is being generated since I don't provide a name?

```
  aws-native:sqs:Queue (StockBotQueue7013873D):
    error: operation CREATE failed with "GeneralServiceException": The name of a FIFO queue can only include alphanumeric characters, hyphens, or underscores, must end with .fifo suffix and be 1 to 80 in length. (Service: Sqs, Status Code: 400, Request ID: a9707e5b-8c75-55bd-986b-22165a0f5288)
```

```json
"StockBotQueue7013873D": {
  "Type": "AWS::SQS::Queue",
  "Properties": {
  "ContentBasedDeduplication": true,
  "DeduplicationScope": "messageGroup",
  "FifoQueue": true,
  "FifoThroughputLimit": "perMessageGroupId"
  },
  "UpdateReplacePolicy": "Delete",
  "DeletionPolicy": "Delete",
  "Metadata": {
  "aws:cdk:path": "example-stock-bot/StockBot/Queue/Resource"
  }
},
```

### Expected Behavior

Pulumi should properly create a SQS FIFO queue with a name ending in `.fifo`

### Steps to reproduce

Create a SQS FIFO queue with an auto-generated name

### Output of `pulumi about`

```
CLI
Version      3.53.1
Go Version   go1.19.5
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.28.0
nodejs  unknown

Host
OS       darwin
Version  13.1
Arch     arm64

This project is written in nodejs: executable='/Users/samgoodwin/.nvm/versions/node/v16.14.2/bin/node' version='v16.14.2'

Current Stack: sam-g/stock-bot/stock-bot

TYPE                 URN
pulumi:pulumi:Stack  urn:pulumi:stock-bot::stock-bot::pulumi:pulumi:Stack::stock-bot-stock-bot


Found no pending operations associated with stock-bot

Backend
Name           pulumi.com
URL            https://app.pulumi.com/sam-g
User           sam-g
Organizations  sam-g
```

### Additional context

_No response_

### Contributing

kpitzen:
Introduces the concept of `NamingTrivia` which is a structured set of information about the semantics of names expected by AWS. Currently, this is most relevant to FIFO queues in SQS, since they must always be suffixed by `.fifo`.  I expect the application logic to evolve over time, but since each resource attribute is handled individually, it may never be the case that we need to compose multiple pieces of trivia.

Resolves #794


examples/aws-native-naming-conventions/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import * as assert from "assert";
+
+// Create an AWS resource (AWS SQS FIFO Queue)
+const queue = new aws.sqs.Queue("my-queue",{
+    fifoQueue: true,
+});
+
+// Export the name of the queue
+export const name = queue.queueName;
+name.apply(name => assert.match(name || "", new RegExp(".*\.fifo")));
```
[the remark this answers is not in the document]
kpitzen: I verified that it ran when I was executing integration tests locally!  I could also see doing the validation on the test harness side, but I'm not super opinionated either way.

provider/pkg/provider/ids.go
```diff
@@ -53,14 +93,24 @@ func getDefaultName(
...
+			if len(namingTrivia.suffix) > 0 {
+				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
+					" Prefix: %[2]q is too large to fix max length constraint of %[3]d"+
+					" with required suffix %[4]q. Please provide a value for %[1]q",
+					sdkName, prefix, autoNamingSpec.MaxLength, namingTrivia.suffix)
+			} else {
+				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
+					" Prefix: %[2]q is too large to fix max length constraint of %[3]d. Please provide a value for %[1]q",
+					sdkName, prefix, autoNamingSpec.MaxLength)
+			}
```
lukehoban, line 113: Kind of an aside - but it's a little bit of a shame how much we "reinvent" across providers for autonaming.  Feels like something where we could push a standard library of autonaming utilities into pulumi/pulumi or elsewhere to apply here and elswhere.

TFbridge has a somewhat rich set of utilities for autonaming that can be applied to do I believe everything done here as special cases.  And AFAICT - nothing in `FromName` there is specific to Terraform - so that pattern could be lifted out and used elsewhere potentially.

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/names.go#L240-L260

kpitzen: Agreed - it's all resource spec / attribute mappings to string manipulation which should just not care how the resources themselves were defined.  I'd love to chat about how to handle this more generally sometime - especially when it comes to provider-specific semantics.

examples/aws-native-naming-conventions/index.ts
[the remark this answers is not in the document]
kpitzen: I went ahead and added validation via `ExtraRuntimeValidation` as well - might as well lock it from both sides.

provider/pkg/provider/ids.go
```diff
@@ -53,14 +93,24 @@ func getDefaultName(
 		randLength = autoNamingSpec.MinLength - len(prefix)
 	}
 
+	// Generate naming trivia for the resource.
+	namingTrivia := createNamingTrivia(sdkName, news)
+
 	maxLength := 0
 	if autoNamingSpec.MaxLength > 0 {
-		left := autoNamingSpec.MaxLength - len(prefix)
+		left := autoNamingSpec.MaxLength - len(prefix) - len(namingTrivia.Suffix)
```
mikhailshilkov, line 101: We have `autoNammingSpec` that contains metadata about auto-naming. My idea would be to add a `Suffix` property to it. The flow would go like this:
- Naming trivia (similar to yours) are defined in the `schema` folder, not `provider`
- Codegen uses that trivia to build `autoNamingSpec` metadata including Min/MaxLength and suffix
- Metadata are written down to `metadata.json`
- Engineers can easily inspect which resources and properties were affected by looking at the JSON diff
- Provider uses the metadata to build names at runtime, without knowing about `namingTrivia`

This flow is used in Azure Native (and AWS Native for min/max length).

What do you think?

provider/pkg/provider/ids.go
```diff
@@ -21,6 +21,46 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
+var NamingTriviaPropMap map[string]NamingTrivia = map[string]NamingTrivia{
+	"queueName": {
+		Suffix: ".fifo",
```
mikhailshilkov, line 26: In addition to making this change in our code, should we report the need for that to AWS and ask them to include this in the CF specs somehow?

kpitzen: That would be _fantastic_.  Currently there's no real support for naming semantics in the spec, and building that into something of a standard would make our lives much easier.

provider/pkg/provider/ids.go
kpitzen: @lukehoban - still working on polishing up some of the code + comments in this PR, but per @mikhailshilkov 's feedback, I've made this a bit more generalized and integrated it into schema generation for this repo.  That feels a lot more generic wrt which native provider is being tinkered with and may be a potential path forward in other places.

provider/cmd/pulumi-resource-aws-native/semantics.json
```diff
@@ -0,0 +1,22 @@
...
+        "aws-native:sqs:Queue": {
+            "cf": "AWS::SQS::Queue",
+            "namingTriviaSpec": 
+            {
+                "queueName": {
+                    "rule": {
+                        "field": "fifoQueue",
+                        "condition": {
+                            "predicate": "equal",
+                            "value": { "V": true }
```
kpitzen, line 12: I went back and forth a bit wrt whether this ought to be an explicit `PropertyValue` in the spec, or if we ought to coerce it into one as we parse the spec.  I'm sort of indifferent, but if there are any strong opinions against this pattern, I'm happy to change it.

provider/pkg/provider/ids.go
```diff
@@ -73,5 +86,11 @@ func getDefaultName(
 	if err != nil {
 		return resource.PropertyValue{}, err
 	}
+
+	// Apply naming trivia to the generated name.
+	if namingTriviaApplies {
+		random = namingTrivia.ApplyTrivia(sdkName, random)
+	}
```
kpitzen, line 93: Sort of a code style question - do we prefer adding explicit branches in the outer layer like this, or would having the `ApplyTrivia` method on `NamingTrivia` be a no-op in the case where we've decided it doesn't apply make more sense?  I tend to prefer not to add hidden complexity when possible, but am open to arguments otherwise.

provider/pkg/schema/semantics.go
```diff
@@ -0,0 +1,112 @@
...
+// about a resource's name that is used to generate a random name.
+type NamingTrivia struct {
+	Suffix string `json:"suffix,omitempty"`
+}
+
+var NamingTriviaPropMap map[string]NamingTrivia = map[string]NamingTrivia{
+	"queueName": {
+		Suffix: ".fifo",
+	},
+}
```
danielrbradley, line 66: Is this a left-over from before adding the semantics to the metadata?

provider/pkg/provider/ids.go
danielrbradley: I think this is fine given this existing structure of the code. There might be some scope to restructure this a little more in the future. Possibly calculating the random length after getting the trivia. Similarly, the validation of the length could be done after putting the whole id together for simplicity - so we could include the example generated id in the error message so the user can see how it's constructed. This isn't really needed to get this feature out though - just making notes while I'm parsing.

danielrbradley (commented):
This looks pretty close - just a question on if there's some left-over code in the semantics.go that needs cleaning up.

provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -73,8 +75,11 @@ func main() {
 
 	supportedTypes := readSupportedResourceTypes(genDir)
 	jsonSchemas := readJsonSchemas(schemaFolder)
-
-	fullSpec, _, err := schema.GatherPackage(supportedTypes, jsonSchemas, true)
+	semanticsDocument, err := schema.GatherSemantics(semanticsDir)
+	if err != nil {
+		panic(err)
```
mikhailshilkov, line 80: Nit: Worth wrapping similar to below?
```suggestion
		panic(fmt.Sprintf("error gathering semantics: %v", err))
```

mikhailshilkov (commented):
This new approach makes sense to me, thank you for refactoring. I'll let you resolve Dan's comments, LGTM otherwise.

provider/pkg/provider/ids.go
```diff
@@ -53,14 +53,27 @@ func getDefaultName(
 		randLength = autoNamingSpec.MinLength - len(prefix)
```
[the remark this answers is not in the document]
kpitzen: Ahh, very good catch - let me fix this up.  In this case, it has less to do with the gap between min and max length, and more to do with the gap between the length of the name and MinLength when the difference between MinLength and MaxLength is small - basically, when we're constrained by the number of random characters we can add.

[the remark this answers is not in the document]
kpitzen: Thinking about this more - I don't think we need to do anything other than including namingTrivia as a part of the current autonaming logic.  If the lengths of prefix and naming trivia are less than the min, we'll only add characters to meet that constraint.  If they exceed the max, we already have escape logic to notify the user we cannot meet the naming constraints via autonaming.  I was overcomplicating it in my head :)

danielrbradley (approved):

PR #818, opened 2023-02-27
Fix conversion of dynamo table resource payloads

### LINKED ISSUE #728: Dynamodb createTable keySchema does not match with specification
### What happened?

Creating a dynamodb table with the AWS native library does not succeed as the content of the keySchema is not converted to uppercase for AWS.

Lib version used: 0.43.0

### Steps to reproduce

```
const table = new aws.dynamodb.Table("table", {
  attributeDefinitions: [
    {
      attributeName: "test",
      attributeType: "S",
    },
    {
      attributeName: "Part",
      attributeType: "S",
    },
  ],
  keySchema: [
    {
      attributeName: "test", // <-- this must be uppercase currently to not get an error on this part.
      keyType: "HASH",
    },
    {
      attributeName: "Part",
      keyType: "RANGE",
    },
  ],
  tableName: "table",
  billingMode: "PAY_PER_REQUEST",
});
```

### Expected Behavior

As the pulumi documentation indicates, `attributeName` and `keyType` should be written as such and not with uppercase.

### Actual Behavior

```
 aws-native:dynamodb:Table (healthcare-providers):
    error: creating resource: operation error CloudControl: CreateResource, https response error StatusCode: 400, RequestID: 60c33dc6-3a65-4bf8-88ed-adbf151f9800, api error ValidationException: Model validation failed (#/KeySchema: #: 0 subschemas matched instead of one
    #/KeySchema/0: required key [KeyType] not found
    #/KeySchema/0: required key [AttributeName] not found
    #/KeySchema/0: extraneous key [attributeName] is not permitted
    #/KeySchema/0: extraneous key [keyType] is not permitted
    #/KeySchema/1: required key [KeyType] not found
    #/KeySchema/1: required key [AttributeName] not found
    #/KeySchema/1: extraneous key [attributeName] is not permitted
    #/KeySchema/1: extraneous key [keyType] is not permitted
    #/KeySchema: expected type: JSONObject, found: JSONArray)
```

### Output of `pulumi about`

```
Dependencies:
NAME                VERSION
@pulumi/aws         5.23.0
@pulumi/aws-native  0.43.0
@pulumi/pulumi      3.48.0
```

### Additional context

When I convert the letters into uppercase, the error disappears.

### Contributing

danielrbradley:
Fixes #728
- Add testing of conversion for real resources from the real metadata.
- Add conversion of OneOf specs.
- Ignore conversions of OneOf specs which result in no change. E.g. Any

### DIFF
--- a/provider/pkg/schema/convert.go
+++ b/provider/pkg/schema/convert.go
@@ -58,6 +58,18 @@ func (c *sdkToCfnConverter) sdkTypedValueToCfn(spec *pschema.TypeSpec, v interfa
 		return c.sdkObjectValueToCfn(typName, v)
 	}
 
+	if spec.OneOf != nil {
+		for _, item := range spec.OneOf {
+			converted := c.sdkTypedValueToCfn(&item, v)
+			// If the conversion resulted in a different value, return it.
+			// We should probably fail if we can't convert rather than considering change a success.
+			// This is likely to happen if one of the types is any.
+			if !reflect.DeepEqual(converted, v) {
+				return converted
+			}
+		}
+	}
+
 	switch spec.Type {
 	case "array":
 		array := v.([]interface{})
--- a/provider/pkg/schema/convert_examples_test.go
+++ b/provider/pkg/schema/convert_examples_test.go
@@ -1,0 +1,50 @@
+// Copyright 2016-2021, Pulumi Corporation.
+
+package schema
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDynamoKeySchemaConversion(t *testing.T) {
+	input := map[string]interface{}{
+		"keySchema": []interface{}{
+			map[string]interface{}{
+				"attributeName": "test",
+				"keyType":       "HASH",
+			},
+			map[string]interface{}{
+				"attributeName": "Part",
+				"keyType":       "RANGE",
+			},
+		},
+	}
+	expected := map[string]interface{}{
+		"KeySchema": []interface{}{
+			map[string]interface{}{
+				"AttributeName": "test",
+				"KeyType":       "HASH",
+			},
+			map[string]interface{}{
+				"AttributeName": "Part",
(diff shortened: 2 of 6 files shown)


provider/pkg/schema/convert.go
```diff
@@ -58,6 +58,18 @@ func (c *sdkToCfnConverter) sdkTypedValueToCfn(spec *pschema.TypeSpec, v interfa
...
 	}
 
+	if spec.OneOf != nil {
+		for _, item := range spec.OneOf {
+			converted := c.sdkTypedValueToCfn(&item, v)
+			// If the conversion resulted in a different value, return it.
+			// We should probably fail if we can't convert rather than considering change a success.
+			// This is likely to happen if one of the types is any.
+			if !reflect.DeepEqual(converted, v) {
+				return converted
```
thomas11, line 68: I find this code a bit surprising. Why return the first successful `OneOf`, what if there are more? And when the conversion results in an identical object, is that not still a valid CFN object?

danielrbradley: Yeah, the whole way this conversion works is questionable. Essentially a failure to convert is just returning the input back as the result.

This was the smallest possible change I could find without reworking all the functions. We would need some way to check if the input is compatible with the target in a "tryConvert" method then return the first successful result. Then again, technically we're using `oneOf` here, not `anyOf` so we should be failing if more than one matches ... but the second case in this `oneOf` is any - so that would always fail.

provider/pkg/schema/convert_examples_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDynamoKeySchemaConversion(t *testing.T) {
```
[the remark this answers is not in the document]
danielrbradley: What's the 2 cases you're referring to? I think this only tests the typed value - the second value is kinda redundant in the test, but are required in an actual deployment.

thomas11 (approved):

PR #909, opened 2023-05-16
Update gradle-build-action to resolve vulnarability

mikhailshilkov:
Resolves https://github.com/pulumi/pulumi-aws-native/security/dependabot/55 and https://github.com/pulumi/pulumi-aws-native/security/dependabot/54


thomas11 (approved):

mikhailshilkov:
Hmm, it looks like I haven't understood the reason correctly. Now we have four alerts instead of two 🤔  https://github.com/pulumi/pulumi-aws-native/security/dependabot

PR #936, opened 2023-06-16
Weekly modules update

danielrbradley:
Fix job command to use `-compat=1.17`
Successful run: https://github.com/pulumi/pulumi-aws-native/actions/runs/5292118737/jobs/9578593711
This updates the generated SDKs:
- dotnet uses 6.0.0 project version
- go to uses `errors` rather than `github.com/pkg/errors`
- updated go provider config parsing
- python requires `>=3.7`


mjeffryes:
Can we close this? I think we ended up bumping the go version to 1.20 in a different PR

danielrbradley:
Yup, think this isn't of value to keep

Issue #950, opened 2023-06-28
[schema/codegen] Schema and SDK generation is non-deterministic


kpitzen:
### What happened?

Today, when generating schema from CloudFormation specifications, there is an amount of non-determinism over each run, which causes the resulting schema to change each time it is generated.  Additionally, from that schema, SDK generation is similarly nondeterministic, so given a fixed schema file, the resulting SDKs can change run to run.  This causes CI to be very fragile, needing to be run several times each release in order to be green.

Note: Currently, this appears to only affect documentation within the schema and SDKs, not the structural type definitions or resource/invoke implementations.

An example of this came up from our nightly SDK generation:
https://github.com/pulumi/pulumi-aws-native/pull/949

This PR is based off the exact same cloudformation specs as the currently released version, but nondeterminism meant this PR was opened erroneously (and subsequently failed CI due to dirty working tree issues, since the following run reverted the change).

### Expected Behavior

Given a fixed set of cloudformation schemata, the resulting Pulumi schema should remain the same run-to-run.

### Steps to reproduce

In `pulumi-aws-native`, run `make generate_schema` several times in succession. 

Similarly, running `make generate_<some-sdk-runtime>` several times in succession will yield similarly unpredictable results.

### Output of `pulumi about`

```shell
CLI          
Version      3.73.0
Go Version   go1.20.5
Go Compiler  gc

Host     
OS       darwin
Version  12.2.1
Arch     arm64

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/kpitzen
User           kpitzen
Organizations  kpitzen, pulumi
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mjeffryes:
I think #952 may have actually fixed this. I can seem to get it to reproduce. Let's close for now and reopen if we see it again?

PR #952, opened 2023-06-29
Don't overwrite existing type names

thomas11:
Quoting the added comment:
> For the fully qualified name, we prepend the resource name to the type name if it's already there. This avoids name collisions of types from different modules. However, we don't do this if it would result in a name that already exists.

This fixes some types potentially having a wrong description, and spurious schema diffs caused by iteration order - in case of a name clash, the second type wins.

~10 repeats of `make generate_schema` succeeded for me locally without creating new changes.

Also adds a missing line break to some log statements which lead to a soup of messages on one line.


thomas11:
This PR contributes to fixing #950 but is likely not the whole story.

kpitzen (approved):

Issue #901, opened 2023-05-06
Schema discovery should remove existing files to avoid preserving removed resources


thomas11:
`make discovery` doesn't clean `aws-cloudformation-schema` before downloading the new schema. This could lead to JSON files   being used that were removed upstream.

kpitzen:
This is a good catch - are we concerned with breaking changes that may arise from this? Do we need a way to manually curate what's removed from the resulting schema when it's removed upstream? What would a successful design look like here?

PR #973, opened 2023-07-12
Clean the aws-cloudformation-schema dir before downloading new schemas

mjeffryes:
Addresses issue #901 by deleting and recreating the aws-cloudformation-schema/ dir before downloading the new schemas.

## Testing
Added a new unit test to cover the directory cleaning behavior; ran `make discovery` manually to observe the impact.

I did not include the files changed as a result of running `make discovery` in this commit, but locally my diff shows two files deleted as a result of running with this change:
```
	deleted:    aws-cloudformation-schema/aws-gamecast-application.json
	deleted:    aws-cloudformation-schema/aws-gamecast-streamgroup.json
```

## Rollout
N/A

## Rollback/Revert
safe to revert

### DIFF
--- a/provider/cmd/pulumi-gen-aws-native/main.go
+++ b/provider/cmd/pulumi-gen-aws-native/main.go
@@ -219,18 +219,14 @@ func cleanDir(dir string, perm os.FileMode) error {
 		return err
 	}
 
-	err = os.MkdirAll(dir, perm)
-	if err != nil {
-		return err
-	}
-	return nil
+	return os.MkdirAll(dir, perm)
 }
 
 func downloadCloudFormationSchemas(urls []string, outDir string) error {
 	// start with an empty directory
 	err := cleanDir(outDir, 0755)
 	if err != nil {
-		panic(err)
+		return err
 	}
 
 	for _, url := range urls {
--- a/provider/cmd/pulumi-gen-aws-native/main.go
+++ b/provider/cmd/pulumi-gen-aws-native/main.go
@@ -213,7 +213,26 @@ func writeSupportedResourceTypes(outDir string) error {
 	return emitFile(outDir, supportedResourcesFile, []byte(supportedContent))
 }
 
+func cleanDir(dir string, perm os.FileMode) error {
+	err := os.RemoveAll(dir)
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(dir, perm)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func downloadCloudFormationSchemas(urls []string, outDir string) error {
+	// start with an empty directory
+	err := cleanDir(outDir, 0755)
+	if err != nil {
+		panic(err)
+	}
+
 	for _, url := range urls {
 		resp, err := http.Get(url)
 		if err != nil {
--- a/provider/cmd/pulumi-gen-aws-native/main_test.go
+++ b/provider/cmd/pulumi-gen-aws-native/main_test.go
@@ -1,0 +1,69 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


mjeffryes:
I don't think this needs a changelog entry since it's only changing a build tool, not actually changing the provider or schemas, but let me know if you disagree.

provider/cmd/pulumi-gen-aws-native/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	file.Close()
+
+	err = cleanDir(dir, 0755)
+	if assert.Nil(err) {
+		assertExistsAndEmpty(t, dir)
+	}
+}
+
+func assertExistsAndEmpty(t *testing.T, path string) bool {
+  assert := assert.New(t)
```
mikhailshilkov, line 57: Nit: something looks off with spacing in this line

provider/cmd/pulumi-gen-aws-native/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if assert.DirExists(path) {
+		dir, err := os.Open(path)
+		if assert.Nil(err) {
+			defer dir.Close()
+
+			_, err = dir.Readdirnames(1)
+			return assert.Equalf(io.EOF, err, "Non-empty dir")
+		}
+	}
+	return false
```
mikhailshilkov, line 68: I think you ignore the return result. Shouldn't it be an assert instead of return, so that the test fails if the dir does not exist?

provider/cmd/pulumi-gen-aws-native/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	err = cleanDir(dir, 0755)
+	if assert.Nil(err) {
+		assertExistsAndEmpty(t, dir)
+	}
+}
+
+func assertExistsAndEmpty(t *testing.T, path string) bool {
+  assert := assert.New(t)
+
+	if assert.DirExists(path) {
```
mikhailshilkov, line 59: Nit: consider flipping the if condition and short-circuiting return/assert.

provider/cmd/pulumi-gen-aws-native/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	temp, err := os.MkdirTemp("", "test_CleanDir")
+	require.Nil(err)
+	t.Cleanup(func() { os.RemoveAll(temp) })
+
+	dir := temp + "/foo"
+
+	// when called on non-existent dir
+	require.NoDirExists(dir)
+	err = cleanDir(dir, 0755)
+	if assert.Nil(err) {
```
mikhailshilkov, line 40: Does `assert.Nil` panic if `err` is not empty? If it does, do we need this `if` statement? Could `assertExistsAndEmpty` be on a new line simply after assert?

mikhailshilkov (commented):
Looks great overall. A few minor comments around error checking and readability.

provider/cmd/pulumi-gen-aws-native/main_test.go
mjeffryes: `require.X` will stop the test execution; `assert.X` records the failure and continues. (This is the reason all the methods on `assert` return bool.) I think the `require` vs `assert` decision is largely stylistic; in this case, I'm using assert b/c it's a property I'm trying to test rather than a pre-condition for the test.

provider/cmd/pulumi-gen-aws-native/main_test.go
mjeffryes: I'm just following the convention from the `stretchr/testify/assert` package that assert methods always return bool so you can use the result to skip parts of a test that depend on that assert.

provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -213,7 +213,26 @@ func writeSupportedResourceTypes(outDir string) error {
...
+		return err
+	}
+	return nil
+}
+
 func downloadCloudFormationSchemas(urls []string, outDir string) error {
+	// start with an empty directory
+	err := cleanDir(outDir, 0755)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
mjeffryes: No strong reason; I just figured this is a blocking error and saw that we're panicking on error elsewhere in this command, but I do see that this particular method is just returning err so I'll change to adopt that

provider/cmd/pulumi-gen-aws-native/main_test.go
mikhailshilkov: TIL! I think I'm getting rusty. Do we use both in other parts of the codebase?

mikhailshilkov (approved):
LGTM

PR #986, opened 2023-07-20
Fix type generation for AnyOf; improve OneOf

### LINKED ISSUE #777: [schema-generation] Strage type resolution on schema generation
### What happened?

AWS CloudFormation Metric Stream property types were changed from:
```json
    "CreationDate" : {
      "oneOf" : [ {
        "format" : "date-time"
      }, {
        "format" : "timestamp"
      } ],
      "description" : "The date of creation of the metric stream.",
      "type" : "string"
    },
```

to:
```json
    "CreationDate" : {
      "description" : "The date of creation of the metric stream.",
      "type" : "string",
      "anyOf" : [ {
        "format" : "date-time"
      }, {
        "format" : "timestamp"
      } ]
```

Which resulted in an output schema type of:
```json
                "creationDate": {
                    "oneOf": [
                        {
                            "$ref": "pulumi.json#/Any"
                        },
                        {
                            "$ref": "pulumi.json#/Any"
                        }
                    ],
                    "description": "The date of creation of the metric stream."
                },
```

### Steps to reproduce

Define a schema object with a "oneOf" key with multiple format specifications.

### Expected Behavior

Expected behavior would be a schema object like this:
```json
                "creationDate": {
                    "type": "string",
                    "description": "The date of creation of the metric stream."
                },
```
or:
```json
                "creationDate": {
                    "type":  { "$ref": "pulumi.json#/Any" },
                    "description": "The date of creation of the metric stream."
                },
```

### Actual Behavior

image

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

mjeffryes:
Improves type generation for CF schemas that use AnyOf or OneOf, by addressing schema factoring and refactoring generation to avoid drift between implemtation for these two fields. 

Fixes #777 

## Description
There are essentially 3 small fixes in this change:

1) CloudFormation schemas are based on [Json Schema](json-schema.org) which supports two fields for declaring variants with multiple arms: `anyOf` and `oneOf`. They are [specified slightly differently](https://json-schema.org/understanding-json-schema/reference/combining.html), but for our purposes, we can map both to the OneOf field of pulumi's TypeSpec.  Unfortunately, we'd lost parity between the generation code for these two as fixes were only applied to one path or the other. This change consolidates the generation for AnyOf and OneOf to reduce drift and bring AnyOf up to parity with OneOf.

2) Json Schema (and CF schemas) also allow [schema factoring](https://json-schema.org/understanding-json-schema/reference/combining.html#factoring-schemas) where fields included on the containing schema are applied to every arm in the anyOf/anyOf variant. This change adds support to correctly generate types from factored schemas. (This is the root fix for #777.)

3) Often the arms of a oneOf/anyOf don't differ in ways that manifest in the pulumi type/schema. (eg. two string variants with different formats as in #777). With this change we deduplicate the types we generate in a OneOf TypeSpec.

### Additional notes
Item 1 above introduces a digit into the type name for objects declared in an AnyOf. This is consistent with OneOf, but feels a bit icky at times b/c many times theres only one arm that's actually an object so the numbering is not required for disambiguation. It might be worth exploring another naming change to improve consistency here 

Reviewers note: I've split the code changes and schema changes into separate commits to improve readability, but plan to squash on merge.

## Testing
There's a severe lack of tests overall in this code, but I added a few units that are relevant to #777 and carefully sanity checked the changes to the schema.

### DIFF
--- a/provider/pkg/schema/gen.go
+++ b/provider/pkg/schema/gen.go
@@ -933,44 +933,34 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
 		return &pschema.TypeSpec{Ref: referencedTypeName}, nil
 	}
 
-	// Union types.
-	if len(propSchema.AnyOf) == 1 {
-		return ctx.propertyTypeSpec(parentName, propSchema.AnyOf[0])
-	}
-
-	if len(propSchema.AnyOf) > 1 {
-		var types []pschema.TypeSpec
-		for _, sch := range propSchema.AnyOf {
-			typ, err := ctx.propertyTypeSpec(parentName, sch)
-			if err != nil {
-				return nil, err
-			}
-			types = append(types, *typ)
-		}
-		return &pschema.TypeSpec{
-			OneOf: types,
-		}, nil
-	}
-
-	if len(propSchema.OneOf) == 1 {
-		return ctx.propertyTypeSpec(parentName, propSchema.OneOf[0])
-	}
-
-	if len(propSchema.OneOf) > 1 {
-		var types []pschema.TypeSpec
-		for i, sch := range propSchema.OneOf {
+	// Union types
+	if len(propSchema.AnyOf) > 0 || len(propSchema.OneOf) > 0 {
+		schemas := FlattenJSSchema(propSchema)
+		types := make([]pschema.TypeSpec, 0, len(schemas))
+		typeMap := make(map[string]bool, len(schemas))
+		for i, sch := range schemas {
 			typ, err := ctx.propertyTypeSpec(fmt.Sprintf("%s%d", parentName, i), sch)
 			if err != nil {
 				return nil, err
 			}
-			types = append(types, *typ)
-		}
-		for _, typ := range types {
-			if typ.Ref != "pulumi.json#/Any" {
-				return &pschema.TypeSpec{
-					OneOf: types,
-				}, nil
+
+			// stringify to create a key for identifying duplicates
+			keybytes, err := json.Marshal(*typ)
+			if err != nil {
+				return nil, err
 			}
+			key := string(keybytes)
+
+			if _, ok := typeMap[key]; !ok {
+				types = append(types, *typ)
+				typeMap[key] = true
+			}
+		}
+
+		if len(types) == 1 {
+			return &types[0], nil
+		} else {
+			return &pschema.TypeSpec{OneOf: types}, nil
 		}
 	}
 
--- a/provider/pkg/schema/gen_examples_test.go
+++ b/provider/pkg/schema/gen_examples_test.go
@@ -1,0 +1,93 @@
+// Copyright 2016-2021, Pulumi Corporation.
+
+package schema
+
+import (
+	//	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	jsschema "github.com/lestrrat-go/jsschema"
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+)
+
+type PropertyTypeSpecTestCase struct {
+	json     string
+	expected pschema.TypeSpec
+}
+
+func TestPropertyTypeSpec(t *testing.T) {
+	cases := []PropertyTypeSpecTestCase{
+		{
+			json: `{
+				"type": "string",
+				"oneOf": [
+					{"format": "date-time"},
+					{"format": "timestamp"}
+				]
+			}`,
+			expected: pschema.TypeSpec{Type: "string"},
+		},
+		{
+			json: `{
+				"type": "string",
+				"anyOf": [
+					{"format": "date-time"},
+					{"format": "timestamp"}
+				]
+			}`,
+			expected: pschema.TypeSpec{Type: "string"},
+		},
+		{
+			json: `{
+				"oneOf": [
+					{"type": "string", "format": "date-time"},
+					{"type": "string", "format": "timestamp"}
+				]
+			}`,
+			expected: pschema.TypeSpec{Type: "string"},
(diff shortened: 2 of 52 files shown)


aq17 (approved):
This LGTM - will let @danielrbradley have the final approval

provider/pkg/schema/jsschema_util.go
```diff
@@ -0,0 +1,203 @@
...
+
+package schema
+
+import (
+	jsschema "github.com/lestrrat-go/jsschema"
+)
+
+// FlattenJSSchema recursively flattens a schema containing AnyOf or OneOf into a list of schemas
+// Note: this only performs a shallow copy, don't use in situations where schema fields may be mutated.
+// TODO: this should probably handle AllOf and Not as well.
```
mikhailshilkov, line 11: Every (new) TODO in our code should reference an issue that explains how we will remove it.

provider/pkg/schema/jsschema_util.go
```diff
@@ -0,0 +1,203 @@
...
+func FlattenJSSchema(sch *jsschema.Schema) jsschema.SchemaList {
```
mikhailshilkov, line 12: Should we add tests for these new functions?

mikhailshilkov (commented):
Took a quick look. Code change, added tests, PR description all look great. I will do a full review on Monday if Dan doesn't beat me to it.

provider/pkg/schema/gen.go
```diff
@@ -933,45 +933,35 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
...
-	if len(propSchema.OneOf) == 1 {
-		return ctx.propertyTypeSpec(parentName, propSchema.OneOf[0])
-	}
 
-	if len(propSchema.OneOf) > 1 {
-		var types []pschema.TypeSpec
-		for i, sch := range propSchema.OneOf {
-			typ, err := ctx.propertyTypeSpec(fmt.Sprintf("%s%d", parentName, i), sch)
+			// stringify to create a key for identifying duplicates
+			keybytes, err := json.Marshal(*typ)
```
iwahbe, line 948: I don't believe that unions have ordered keys, so this could allow duplicates to slip through. I will let @danielrbradley decide if that matters.

provider/pkg/schema/gen_examples_test.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2016-2021, Pulumi Corporation.
+
+package schema
+
+import (
+	//	"encoding/json"
```
iwahbe, line 6: Remove left over comment.

provider/pkg/schema/gen_examples_test.go
```diff
@@ -0,0 +1,93 @@
...
+				]
+			}`,
+			expected: pschema.TypeSpec{
+				OneOf: []pschema.TypeSpec{{Type: "string"}, {Type: "number"}},
+			},
+		},
+	}
+
+	for _, tt := range cases {
+		// t.Log("JSON: ", tt.json) // uncomment to debug
```
iwahbe, line 85: Remove left over comment.

provider/cmd/cf2pulumi/schema-full.json
```diff
@@ -149549,14 +149545,7 @@
...
                 "lastUpdateDate": {
-                    "oneOf": [
-                        {
-                            "$ref": "pulumi.json#/Any"
-                        },
-                        {
-                            "$ref": "pulumi.json#/Any"
-                        }
-                    ],
+                    "type": "string",
```
iwahbe, line 149548: This is a breaking change, even if its a very positive one. Again, I'll leave to @danielrbradley for a final call.

provider/cmd/cf2pulumi/schema-full.json
```diff
@@ -35477,7 +35477,8 @@
             },
             "type": "object"
         },
-        "aws-native:ecs:ClusterCapacityProviderAssociationsCapacityProvider": {
+        "aws-native:ecs:ClusterCapacityProviderAssociationsCapacityProvider0": {
```
iwahbe, line 35480: I don't see a `aws-native:ecs:ClusterCapacityProviderAssociationsCapacityProvider1`. If we don't have more then one variant, can we remove the trailing zero?

iwahbe (commented):
Looks like a large improvement. I left a couple of comments on things that stood out to me.

Thanks for separating the PR into logically coherent commits! It made review much easier.

provider/cmd/cf2pulumi/schema-full.json
mjeffryes: Yeah, the other arm here is a primitive type so there isn't any `aws-native:ecs:ClusterCapacityProviderAssociationsCapacityProvider1` generated. Unfortunately, with the way the object types are generated, we have to add the digit before we know how many object types will be generated. I have one idea that might work for these simple cases though.

provider/pkg/schema/jsschema_util.go
mjeffryes: These are a little bit of a hassle to test; I've tested the cases we care about most in gen_examples_test, but I'll add some property tests for these two to boost coverage.

provider/cmd/cf2pulumi/schema-full.json
mjeffryes: I managed to avoid using the ugly type names by detecting this case and regenerating. We still generate the names though, so they clutter the schema a bit. I filed a TODO to track avoiding generating them to begin with. (I suspect it may require some pretty invasive changes to gen though.)

mikhailshilkov (approved):
Took a deeper look and it LGTM

PR #990, opened 2023-07-21
RFC: Use devbox to define a consistent dev environment for contributors

mjeffryes:
Adds `devbox.json` and `.devcontainer` files to define a reproducible build environment. Updated the README with instructions for use.

Devbox orchestrates a nix shell with a specified dependency list, which you can use directly, or inside a container. devcontainer provides a standard for editors and other tools to work inside in a container.

I personally like this approach as a relatively lightweight way to maintain a known good build environment; I'm curious to get the team's feedback. 

Ideally, we'd drive our CI off the same docker image for consistency between local builds and CI (github has some tooling to support this already: https://github.com/devcontainers/ci); I can add this if we decide to go down this path.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,33 @@
+FROM debian:stable-slim
+
+# Step 1: Installing dependencies
+RUN apt-get update
+RUN apt-get -y install bash binutils git gnupg2 xz-utils wget sudo
+
+# Step 2: Setting up devbox user
+ENV DEVBOX_USER=devbox
+RUN adduser $DEVBOX_USER
+RUN usermod -aG sudo $DEVBOX_USER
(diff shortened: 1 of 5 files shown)


danielrbradley (commented):
Really like this! Can we also set up automated updates for this so the tool versions don't go stale?

t0yv0:
Do it! I'm a Nix user (but avoiding containers as much as possible) and can help keep things maintained at the Nix layer. I've not used Devbox but OK to learn it if it gives devcontainers seamlessly which has been requested a few times.

mjeffryes:
> Really like this! Can we also set up automated updates for this so the tool versions don't go stale?

Yes, I don't know anything about our existing automation tooling, but should be straightforward to have a job to run a devbox update, and send a PR.

mikhailshilkov:
Do you want to open follow-up issues for the ideas that came up in comments?

Issue #850, opened 2023-03-29
rds.DBCluster members have non-idiomatic names


jkodroff:
### What happened?

The names of inputs for rds.DBCluster probably need to be aliased to more idiomatic names: https://www.pulumi.com/registry/packages/aws-native/api-docs/rds/dbcluster/

In TS, we have e.g. [dBClusterIdentifier](https://www.pulumi.com/registry/packages/aws-native/api-docs/rds/dbcluster/#dbclusteridentifier_nodejs) (should probably be `dbClusterIdentifier`).

In Python, we have `d_b_cluster_identifier`. That's definitely wrong.

### Expected Behavior

n/a

### Steps to reproduce

n/a

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mjeffryes:
This issue looks like an instance of https://github.com/pulumi/pulumi-aws-native/issues/650

mjeffryes:
resolved by #999

PR #999, opened 2023-07-26
Lowercase initialisms in resource/type names

### LINKED ISSUE #650: aws-native schema mangles capitalized acronyms at start of property names
## Hello!

## Issue details

The name re-styling done to properties doesn't seem to properly grok acronyms at the start of property names (which are represented in all caps in Cloudformation's camel case style). [For example](https://www.pulumi.com/registry/packages/aws-native/api-docs/networkfirewall/rulegroup/#rulegrouprulevariables) `IPSets` becomes

* `iPSets` in Typescript, Java, and YAML _(should be `ipSets`)_
* `i_p_sets` in Python _(should be `ip_sets`)_

grepping around the python SDK for strings like `i_p_`, `a_c_l`, `h_a_` shows this to be a problem on multiple resources. Including [this wonderful example](https://www.pulumi.com/registry/packages/aws-native/api-docs/memorydb/acl/#a_cl_name_python) in `memorydb` where `ACLName` somehow first gets mangled to `aCLName` before being regularized into Python resulting in `a_cl_name`.

As demonstrated by that example though, runs of capitals elsewhere in property names seem to be handled correctly as single words. Though, it seems like there might still be some other bugs related to camelCase parsing in other bits of codegen like [for example](https://github.com/pulumi/pulumi-aws-native/blob/1ae4a4d9c2256b2a79ca536f8d8497b28d10e4c3/sdk/python/pulumi_aws_native/wafv2/get_web_aclassociation.py#L32) `get_web_aclassociation()` instead of `get_web_acl_association()`.

mjeffryes:
Identifies all uppercase acronyms in the CloudFormation Resource names and lowercases them in the generated pulumi schema. This change provides greater consistency in the names we generate in the SDKs and eliminates the confusing misplaced word breaks in python sdks for names beginning with an initialism.

Fixes #650, #850 

# Description
AWS CloudFormation Resources are inconsistent in how they capitalize initialisms. (eg. We see `ResourceARN`, `ServiceAccountRoleArn` and even `NotificationARNs` in different places in the CF schemas) The code generating pulumi schemas currently does a naive downcase of the first character of the names leading to some awkward misplaced wordbreaks in some of our APIs when an name starts with an uppercase initialism (eg. `ACLName` -> `aCLName` -> python: `a_cl_name`). We have a few options for how to resolve this:

1) Convert all initialisms to lowercase (`ACLName ` -> `aclName`, `ResourceARN` -> `resourceArn`, `ServiceAccountRoleArn` -> `serviceAccountRoleArn`)
>   **Pros**
>  - Initialisms have consistent capitalization in pulumi APIs
>  - Can be done in aws-native without coordination with codegen.

>  **Cons**
>  - Casing is different from CF docs. 
>  - We need to add metadata to the provider to reconstruct the CF name
>  - Affects a large number of resources

2) Follow CF capitalization by using a new naming convention in pulumi schema  (`ACLName ` -> `ACL_name`, `ResourceARN` -> `resource_ARN`, `ServiceAccountRoleArn` -> `service_account_role_arn`)
>   **Pros**
>  - Capitalization in SDKs follow CF docs and we can reconstruct the CF name from the pulumi name.

>  **Cons**
>  - Requires changes in codegen for all languages
>  - Affects a large number of resources

3) Just lowercase initialisms at the beginning of names  (`ACLName ` -> `ACL_name`, `ResourceARN` -> `resource_ARN`, `ServiceAccountRoleArn` -> `service_account_role_arn`)
>   **Pros**
>  - Can be done in aws-native without coordination with codegen.
>  - Only changes resources that have mangled names like those in #650 and #850

>  **Cons**
>  - Casing is inconsistent AND is different from CF docs 
>  - We need to add metadata to the provider to reconstruct the CF name

This PR adopts approach 1 on the reasoning that both option 1 and option 3 are significantly faster to implement than option 2, but option 1 is a much better UX than option 3 long-term and we'd rather eat the breaking changes now than be stuck with a clunky naming convention forever.
 
# Testing
Added new unittests around naming; spot-checked schema changes, manual testing with the EC2 example to check that the provider is able to reconstruct the correct CF names.


mjeffryes:
## Reviewer Notes
- Recommend reviewing by commit as commit history splits the code changes from the schema/sdk generation
- Whatever solution we take to resolve 650 establishes a naming convention for aws-native resources for the 1.0 and beyond; let's be sure this is convention we can live with for the long-term. 
- Additional discussion [in slack](https://pulumi.slack.com/archives/C011EMDQ8JE/p1689712918701649) and on #650. 

Personally, I don't think doing option 3 makes much sense, but I could be convinced that we should take option 2 if we think the investment in new naming convention for codegen is justified.

iwahbe (approved):
LGTM. I really appreciate the degree of round trip testing.

provider/cmd/pulumi-resource-aws-native/schema.json
```diff
@@ -5273,12 +5273,12 @@
...
                     "$ref": "#/types/aws-native:appflow:FlowS3SourceProperties"
                 },
-                "sAPOData": {
-                    "$ref": "#/types/aws-native:appflow:FlowSAPODataSourceProperties"
-                },
                 "salesforce": {
                     "$ref": "#/types/aws-native:appflow:FlowSalesforceSourceProperties"
                 },
+                "sapoData": {
+                    "$ref": "#/types/aws-native:appflow:FlowSAPODataSourceProperties"
```
mikhailshilkov, line 5280: Should this change apply to type names as well? `SapoData` vs `FlowSAPODataSourceProperties` looks inconsistent?

provider/cmd/pulumi-resource-aws-native/schema.json
```diff
@@ -46525,28 +46525,28 @@
         },
         "aws-native:kinesisfirehose:DeliveryStreamKMSEncryptionConfig": {
             "properties": {
-                "aWSKMSKeyARN": {
+                "awskmsKeyArn": {
```
mikhailshilkov, line 46528: Another instance of above: `awsKmsKeyArn` would be ideal.

provider/cmd/pulumi-resource-aws-native/schema.json
```diff
@@ -46737,10 +46737,10 @@
             },
             "type": "object",
             "required": [
-                "clusterJDBCURL",
+                "clusterJdbcurl",
```
mikhailshilkov, line 46740: JdbcUrl

mikhailshilkov (commented):
I reviewed the API shape changes. Overall I'm happy with the changes, it's a clear improvement, and I think it's good as the long term approach. Tagging @mnlumi and @lukehoban to take a look, because this is going to stay ~forever.

My biggest question was about type names - see below.

mnlumi:
@mjeffryes and I chatted about this and I am agreement with option 1 for a consistent Pulumi user experience. While it  would be nice to match CF, it seems like a very heavy lift and may still cause confusion in user experience when using this with other providers.

provider/cmd/pulumi-resource-aws-native/schema.json
```diff
@@ -6214,21 +6214,21 @@
         "aws-native:applicationinsights:ApplicationHANAPrometheusExporter": {
             "description": "The HANA DB Prometheus Exporter settings.",
             "properties": {
-                "agreeToInstallHANADBClient": {
+                "agreeToInstallHanadbClient": {
```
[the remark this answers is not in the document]
mjeffryes: I'm pretty sure it's not possible without a human maintained list of acronyms (which is something I don't think we'd want to try to maintain.

provider/cmd/pulumi-resource-aws-native/schema.json
mjeffryes: hmmm... yes, I think it should. I don't think we transform the type names at all today, but for consistency, we should.

mjeffryes: Since this review is already pretty big and is sufficient to resolve the linked issues, I'd like to ship this and then get the type names in a follow up. I filed https://github.com/pulumi/pulumi-aws-native/issues/1008 to track lowercasing initialisms in type names

cjschuck:
This PR has caused us some grief in relation to the function_arn output of cloudfront functions.

`aws-native:cloudfront:Distribution (function-name):

    error: operation UPDATE failed with "InvalidRequest": Invalid request provided: 1 validation error detected: Value null at 'distributionConfig.defaultCacheBehavior.functionAssociations.items.1.member.functionARN' failed to satisfy constraint: Member must not be null (Service: CloudFront, Status Code: 400, Request ID: a60c51c2-f189-486e-8320-c9ea9c1206ee)
`

In place of function_arn returning the ARN of the function, it's returning "output<string>".

Issue #881, opened 2023-04-14
Unable to specify AWS::Pipes::Pipe SelfManagedKafkaAccessConfigurationCredentials due to missing properties


ssemyonov:
### What happened?

According to [AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-selfmanagedkafkaaccessconfigurationcredentials.html) `SelfManagedKafkaAccessConfigurationCredentials` should allow setting one of four authentication options, however corresponding provider class - [PipeSelfManagedKafkaAccessConfigurationCredentialsArgs](https://github.com/pulumi/pulumi-aws-native/blob/master/sdk/dotnet/Pipes/Inputs/PipeSelfManagedKafkaAccessConfigurationCredentialsArgs.cs) for .NET SDK doesn't contain any properties to do that.
When running `pulumi up` with no parameters set for the class above, AWS returns validation error:

```
  #/SourceParameters/SelfManagedKafkaParameters/Credentials: #: 0 subschemas matched instead of one
    #/SourceParameters/SelfManagedKafkaParameters/Credentials: required key [BasicAuth] not found
    #/SourceParameters/SelfManagedKafkaParameters/Credentials: required key [SaslScram512Auth] not found
    #/SourceParameters/SelfManagedKafkaParameters/Credentials: required key [SaslScram256Auth] not found
    #/SourceParameters/SelfManagedKafkaParameters/Credentials: required key [ClientCertificateTlsAuth] not found)

```

### Expected Behavior

The properties in `PipeSelfManagedKafkaAccessConfigurationCredentialsArgs` class match the ones documented in [AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-selfmanagedkafkaaccessconfigurationcredentials.html) and include the following:

```
  BasicAuth: String
  ClientCertificateTlsAuth: String
  SaslScram256Auth: String
  SaslScram512Auth: String
```

### Steps to reproduce

Instantiate an object of a `PipeSelfManagedKafkaAccessConfigurationCredentialsArgs` type in .NET and attempt to set any properties in the code.

### Output of `pulumi about`

```
CLI
Version      3.63.0
Go Version   go1.20.2
Go Compiler  gc

Plugins
NAME        VERSION
aws         5.35.0
aws-native  0.57.0
dotnet      unknown

Host
OS       Microsoft Windows 10 Pro
[...]

PR #1007, opened 2023-07-28
Fix generation of type definitions using OneOf

mjeffryes:
The logic for detecting object types when traversing a ref in the schema was too aggressive; it was eating types using OneOf/AllOf and turning them into empty objects. With this change, it only traps objects that are explicitly typed as "object" or that have properties

Fixes: #881 

## Testing
Added some new examples cases to the unittests, scanned the schema changes


mjeffryes:
The schema-tools report isn't really doing a good job with this diff. The types that are "missing" are where we were mistakenly creating objects. The inputs/outputs that changed to "" actually changed to `OneOf`s

There are a couple cases where types changed to "pulumi.json#/Any". I checked each of these and they all use "propertyPatterns"; it looks like jsschema is not able to parse the Regex on a handful of these.

iwahbe (commented):
Are we assuming that `typeSchema.Type[0]` is representative of all type schemas?

mjeffryes:
> Are we assuming that `typeSchema.Type[0]` is representative of all type schemas?

It's not (That's why we had this issue to begin with: OneOf types often don't have a `type` field at the top level), but we don't need to unambiguously identify all the different types here; we just need to identify things that are "object" shaped  (mostly so we can avoid tacking "Properties" onto the name, which is what happens when we recurse). Objects are identified by schemas that have `type == "object"` or have non-empty `properties` or non-empty `propertiesPattern`.

To be fair, I don't love the way this code is structured; it's trying to do everything in one pass: convert JSSchema types into pulumi types, rename everything and introduce references for intermediate objects. I think we could make the whole thing easier to reason about and test if we split it into at least two passes, but I didn't feel like I had the time to do that right now.

iwahbe (approved):

mikhailshilkov (approved):
LGTM, that was surprisingly straightforward.

Issue #1009, opened 2023-07-28
Fix unsafe type transformations in SdkToCfn


mjeffryes:
As noted in https://github.com/pulumi/pulumi-aws-native/issues/846, the transformation functions in SdkToCfn blindly assume that the value matches the expected type and can panic if it doesn't. In particular, we try to convert values with a oneOf type by blindly trying the first arm. https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/schema/convert.go#L61 

These methods should probably guard and return errors instead.

mjeffryes:
Looks like this panic also shows up in https://github.com/pulumi/pulumi-aws-native/issues/399. (But fixing the panic is not enough to resolve the whole bug for that one either.)

PR #1015, opened 2023-08-01
Add error handling to CfnToSdk

mjeffryes:
Adds basic type checking errors to the internals of CfnToSdk to enable safe conversion of values
of OneOf types.

Fixes: #1009 

## Description
The CfnToSdk conversion code for values lacked any error handling and simply assumed that the value
it was handed matched the expected type. This was a problem for the conversion of values with a 
"OneOf" type as the code blindly tried to convert the value with each type in order, and
which would panic when handed a value for a later arm.

This code introduces some sanity checking around the values and uses the errors to safely find an 
arm that matches the value given.

## Testing
This code should probably get a larger suite of round trip tests but for now I just added a few cases
for basic error handling and the specific situation with oneOf types that was causing the panics


thomas11 (approved):
LGTM!

Issue #1008, opened 2023-07-28
Lowercase initialisms in type names for consistency with resource names


mjeffryes:
As @mikhailshilkov noted in comments on https://github.com/pulumi/pulumi-aws-native/pull/999; that change addresses the awkward resource names reported in #650 and #850, but it doesn't actually lowercase the initialisms in typenames. For consistency, we should lowercase type names as well

PR #1018, opened 2023-08-02
lowercase initialisms in type names for consistency with resource names

mjeffryes:
In #999 we began lowercasing initialisms in property names to prevent erroneous word breaks in the python APIs. This change extends that behavior to all names in the API including resource names, type names and C# module names. 

Closes: #1008

## Description
I worked my way through the schema and lowercased every place that an initialism appeared in all capitals. There were a surprisingly large number of different paths that these names follow (from only a couple sources: Resource names, named type Definitions, and Enum values).

I kind of hate that I had to spread the invocations of `lowercaseAcronyms` all over the place; this is yet another instance where I think the code could be cleaner if we separated naming into a separate pass, but I don't want to delay this change and have it separated too much from the renames in #999.

## Testing

Provider units + manually tested by deploying an `ec2.DhcpOptions` (formerly `ec2.DHCPOptions`) resource.


mjeffryes:
Reviewer note: commits split code change from schema/sdk

mikhailshilkov (approved):
LGTM when you fix the name in the test

PR #1053, opened 2023-08-25
Mark createOnlyProperties with replaceOnChange

### LINKED ISSUE #736: Automatically replace resources when needed
CloudFormation schemas identify certain properties as `createOnly`, meaning they cannot be updated on an existing resource, and require the resource to be re-created to be changed. These are specified in the `createOnlyProperties` field.

This should be used when generating our Schema to correctly mark the inputs as such (`willReplaceOnChanges`).

Resolving this will likely fix the following issues:
- https://github.com/pulumi/pulumi-aws-native/issues/249
- https://github.com/pulumi/pulumi-aws-native/issues/628
- https://github.com/pulumi/pulumi-aws-native/issues/716
- https://github.com/pulumi/pulumi-aws-native/issues/733
- https://github.com/pulumi/pulumi-aws-native/issues/612
- https://github.com/pulumi/pulumi-aws-native/issues/584
- https://github.com/pulumi/pulumi-aws-native/issues/823
- https://github.com/pulumi/pulumi-aws-native/issues/822
- https://github.com/pulumi/pulumi-aws-native/issues/834
- https://github.com/pulumi/pulumi-aws-native/issues/971

mjeffryes:
Updated gen to add the `repalceOnChanges` flag to any properties identified in the CF schema as `createOnlyProperties`.

Fixes #736 (mostly; we'll still need to address https://github.com/pulumi/pulumi-java/issues/1189 to get the correct behavior in java)

## Details

I've implemented this as a post-processing step on the Resource spec to avoid cluttering the already complicated type generation and naming logic.

Ironically, while both the aws CF schema and the sdks use a list of path strings on the root of the resource to designate createOnlyProperties/replaceOnChanges, the schema layer uses a flag on the leaf propertySpecs. So we have to traverse the  structure when setting replaceOnChanges, and then the sdk generation process has to traverse it again to rebuild the paths. 

## Testing
I've added new unit tests and manually verified the schema changes.


provider/pkg/schema/gen_examples_test.go
```diff
@@ -263,3 +265,178 @@ func TestTypeName(t *testing.T) {
...
+		},
+		{
+			createOnlyProperties: codegen.NewStringSet("arr/bar"),
+			resourceSpec:         newBaseResourceSpec(),
+			types:                newBaseTypes(),
+			expectedResource:     newBaseResourceSpec(),
+			expectedTypes:        typesWithReplaceOnChangesBar,
+		},
+		{
+			createOnlyProperties: codegen.NewStringSet("arr/*/bar"),
```
t0yv0, line 402: It seems that '*' is effectively ignored, why is that?

t0yv0:
@mjeffryes the code and the tests look fine, I wonder though if the path-to-schema-edit translation is giving the desired results. It's hard to quite be sure from just the unit tests. There's a number of use cases listed:

https://github.com/pulumi/pulumi-aws-native/issues/736

We're not checking those at integration level? Is there a way to spot-check that this PR changes the schema for these in a way that makes sense? Did you perhaps do that already? as in "manually verified the schema changes."

t0yv0:
I wonder if we need to add schema-tools modification to show these changes because the current check states there are no changes.

t0yv0 (approved):

mjeffryes:
> I wonder if we need to add schema-tools modification to show these changes because the current check states there are no changes.

Aws-native (and all the native providers) are on an older version of the schema check action. It's possible that replaceOnChanges is tracked in the latest version.
> There's a number of use cases listed: https://github.com/pulumi/pulumi-aws-native/issues/736

> We're not checking those at integration level? Is there a way to spot-check that this PR changes the schema for these in a way that makes sense? Did you perhaps do that already? as in "manually verified the schema changes."

I'm open to adding one or two of these as an integration test (though I might need some handholding to figure out how to make programtest do the right thing). I don't think it would make sense to add all the cases identified in #736 as integration tests. 

For manual verification, I found examples in the input schema for `createOnlyProperties` with the following shapes: 

/properties/foo
/properties/foo/bar // where foo might be an object or an []object
/properties/foo/*/bar

and verified that they were faithfully translated into the expected `replaceOnChanges` in the pulumi schema. I also checked the generated sdks in all languages for these examples.

t0yv0:
Perfect. IF you've verified the schema let's move forward then thanks!

provider/pkg/schema/gen_examples_test.go
mjeffryes: The CF schemas are inconsistent about how they identify properties within arrays eg. given the following types:

```
resource Foo {
  bars: []Bar
}

type Bar {
  baz: String
  quux: String
}
```

Foo could have a 
`createOnlyProperties=["/properties/bars/*/baz"]`
or a 
`createOnlyProperties=["/properties/bars/baz"]`

to indicate that the values of baz within bars may only be set on create.(Both variants occur in the latest CF schema.)

We also handle the case 
`createOnlyProperties=["/properties/bars/*"]`
and treat it the same as 
`createOnlyProperties=["/properties/bars"]`.
I did not see any instances of the former in the schema, but it seemed like a reasonable interpretation if it should occur.

I'm happy to add a comment to this effect in the tests.

Issue #1219, opened 2023-12-15
Add support for AWS operation status of `PENDING` 


EronWright:
CI tests are failing with:
``` 
   aws-native:iam:Role (role-39):
      error: unknown status "PENDING": &{ErrorCode: EventTime:2023-12-15 04:58:06.427 +0000 UTC Identifier:<nil> Operation: OperationStatus:PENDING RequestToken:0xc0014ea1d0 ResourceModel:<nil> RetryAfter:<nil> StatusMessage:<nil> TypeName:0xc0014ea1e0 noSmithyDocumentSerde:{}}
```

Root cause seems to be:

https://github.com/pulumi/pulumi-aws-native/blob/a83751e8b5c682541c49420872e42a6063a83c7b/provider/pkg/provider/provider.go#L1135C23-L1169

PR #1228, opened 2023-12-19
Handle PENDING Operation Status  

mjeffryes:
Fixes: https://github.com/pulumi/pulumi-aws-native/issues/1219

This could have been a one liner, but I decided to pull out a helper method to add a test. There are a couple other values of `OperationStatus` that we may need to handle in the future, but I'm punting on those, b/c I haven't seen them in the wild and I'm not 100% sure how they ought to be handled.


danielrbradley (approved):
Looks good 👍 

Just need to get the tests passing!

PR #1283, opened 2024-01-20
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `::set-output` to `"$GITHUB_OUTPUT"`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/nightly-sdk-generation.yml
+++ b/.github/workflows/nightly-sdk-generation.yml
@@ -83,7 +83,7 @@ jobs:
       run: make local_generate
     - name: Git submodule commit hash
       id: vars
-      run: echo ::set-output name=commit-hash::$(git rev-parse HEAD)
+      run: echo commit-hash=$(git rev-parse HEAD) >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


mjeffryes:
Thanks for the PR @arunsathiya! We actually generate these workflows so we'll need to modify the generator or this change will be overridden. See https://github.com/pulumi/ci-mgmt/blob/93512a7493081cb05a8329cf397710f1c8e38a8d/native-provider-ci/src/steps.ts#L65 Looks like there are potentially a few other places we're using `set-output` in there.

arunsathiya:
Thanks for the direction, @mjeffryes! I have a new PR up there: https://github.com/pulumi/ci-mgmt/pull/789 👍🏼

PR #1292, opened 2024-01-25
Add upgrade tests

danielrbradley:
- Set up initial upgrade testing with 2 core resources.
- Use provider attach to make provider debuggable during integration tests.
- Use go:embed instead of printing out source files containing byte arrays.
- Tweak generation to emit ready-compressed files to "./bin" for use within testing.

Fixes https://github.com/pulumi/pulumi-aws-native/issues/114 - so we can run upgrade preview unit tests without credentials in CI.


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -106,7 +108,12 @@ func main() {
 				panic(fmt.Sprintf("error generating examples: %v", err))
 			}
 			providerDir := filepath.Join(".", "provider", "cmd", "pulumi-resource-aws-native")
-			writePulumiSchema(*supportedSpec, providerDir, "schema.json")
+			if err = writePulumiSchema(*supportedSpec, providerDir, "schema.json"); err != nil {
+				break
```
mjeffryes, line 112: Shouldn't this (and the error cases added here) notify in some when writing fails?

provider/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/pulumi/pulumi-aws-native/provider
 
-go 1.21
+go 1.21.0
```
mjeffryes, line 3: curious, why pin to 1.21.0?

provider/go.mod
```diff
@@ -183,28 +188,33 @@ require (
...
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/frand v1.4.2 // indirect
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20211028080628-e2786a622600 // indirect
 )
 
 replace github.com/lestrrat-go/jsschema => github.com/mikhailshilkov/jsschema v0.0.0-20210924145243-fc93fd28ee1b
+
+replace github.com/pulumi/pulumi/pkg/v3 => github.com/pulumi/pulumi/pkg/v3 v3.91.1
+
+replace github.com/pulumi/pulumi/sdk/v3 => github.com/pulumi/pulumi/sdk/v3 v3.91.1
```
mjeffryes, line 220: I'm confused about this too

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+}
+
+func testUpgradeFrom(test *pulumitest.PulumiTest, version string) {
+	test.T().Helper()
+	result := providertest.PreviewProviderUpgrade(test, "aws-native", version)
+	assertpreview.HasNoChanges(test.T(), result)
+}
+
+func testLifecycle(test *pulumitest.PulumiTest) auto.OutputMap {
```
mjeffryes, line 40: this method is unused for now?

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/pulumi/providertest/pulumitest"
+	"github.com/pulumi/providertest/pulumitest/assertpreview"
+	"github.com/pulumi/providertest/pulumitest/assertrefresh"
+	"github.com/pulumi/providertest/pulumitest/opttest"
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/provider"
+	"github.com/pulumi/pulumi/sdk/v3/go/auto"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+)
+
+func TestE2eSnapshots(t *testing.T) {
```
mjeffryes, line 20: love how straightforward this is. :)

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -0,0 +1,82 @@
...
+	t.Run("webAcl-0.94.0", func(t *testing.T) {
+		t.Parallel()
+		test := newAwsTest(t, filepath.Join("testdata", "webAcl"))
+		testUpgradeFrom(test, "0.94.0")
+	})
+}
+
+func testUpgradeFrom(test *pulumitest.PulumiTest, version string) {
+	test.T().Helper()
+	result := providertest.PreviewProviderUpgrade(test, "aws-native", version)
```
mjeffryes, line 36: does this helper include all the logic to record state too?

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		metadataBytes, err := os.ReadFile(filepath.Join(cmdDir, "metadata.json.gz"))
+		if err != nil {
+			return nil, err
+		}
+		return provider.NewAwsNativeProvider(nil, "aws-native", "0.1.0", schemaBytes, metadataBytes)
+	})
+}
+
+func skipIfShort(t *testing.T) {
```
mjeffryes, line 77: noting that this is also unused for now

mjeffryes (approved):
A few notes, but nothing blocking once we get the credentials working so tests can pass

provider/go.mod
danielrbradley: I didn't make this change - I think this was done by `go mod tidy` 🤷

provider/pkg/provider/provider_2e2_test.go
danielrbradley: Yup - there's options to control the state storage but the defaults will store it in the same place as the old version of upgrade test.

provider/cmd/pulumi-gen-aws-native/main.go
danielrbradley: It does, but the existing code pattern is just a bit weird. When it breaks the loop there's then a check for the error. I don't like the pattern, but just keeping it matching the code around it.

provider/go.mod
danielrbradley: When pulling in github.com/pulumi/providertest it seems to force pu/pkg to upgrade, however there's a codegen bug that leave the Go SDK uncompilable. Therefore, we have to use a replacement until the regression is fixed to forcably downgrade pulumi/pkg.

provider/pkg/provider/provider_2e2_test.go
danielrbradley: Oh, here's the docstring for reference:
> PreviewProviderUpgrade captures the state of a stack from a baseline provider configuration, then previews the stack with the current provider configuration. Uses a default cache directory of "testdata/recorded/TestProviderUpgrade/{programName}/{baselineVersion}".

provider/pkg/provider/provider_2e2_test.go
danielrbradley: Yes, until we re-instate the new style integration tests. Will tidy

provider/pkg/provider/provider_2e2_test.go
danielrbradley: Yup, will clean up for now

danielrbradley:
@mjeffryes I think that's all comments addressed. If you can just confirm you're happy with [implement skipping credentials validation](https://github.com/pulumi/pulumi-aws-native/pull/1292/commits/bc785b04f00ec7fdb4071fafa55ec51ee6cf139c) then we're good to go!

provider/go.mod
danielrbradley: New version just out so have upgraded and removed this again 🎉

PR #1326, opened 2024-02-06
Implement skipping credentials validation

### LINKED ISSUE #114: [provider] Implement skipCredentialsValidation

danielrbradley:
Fixes #114 

This allows us to perform previews without needing to have credentials configured.

Side note - it seems as if the not implemented validation for "skip..." are wrong too as they only trigger if the value is set to a falsy value i.e. don't skip. They don't trigger when "skip" is set to true even though the skip is not actually implemented.

Extracted from #1292


mjeffryes (approved):
Any way we could add a unittest for this? I'm OK with skipping if it's too complicated since this is implicitly tested by the upgrade tests your adding, but always nice when we can cover our changes with a unittest.

danielrbradley:
Yes, I think doing this via the upgrade tests will do for now but I'll take this under advisement for as we implement more of these configuration settings.

PR #1342, opened 2024-02-15
Fix map types

danielrbradley:
- Check for AdditionalProperties before falling through to a generic "object" type.
- If PatternProperties only has a single case, treat it like a normal Map type. We could improve this in the future by capturing the regex pattern for documentation and preview-time validation.


provider/pkg/schema/gen.go
```diff
@@ -1092,6 +1092,27 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
 		}
 	}
 
+	var mapType *jsschema.Schema
+	if propSchema.AdditionalProperties != nil {
+		mapType = propSchema.AdditionalProperties.Schema
+	} else if propSchema.PatternProperties != nil && len(propSchema.PatternProperties) == 1 {
+		// TODO: Capture pattern add add to documentation or even provider metadata for early validation feedback.
```
mjeffryes, line 1099: Let's file an enhancement ticket for this.

mjeffryes (approved):
Nice, that fixes a lot of types! Can you add a few test cases to gen_examples_test as well?

mjeffryes:
This should let us close a few of the open schemagen issues.

provider/pkg/schema/gen.go
danielrbradley: Good shout. Added this: https://github.com/pulumi/pulumi-aws-native/issues/1346

PR #1347, opened 2024-02-16
Add "unexpected tags shapes" report

danielrbradley:
Identify which style of tags each resource has and list unidentified schemas in a report to help us address them iteratively.

Once the tags work is complete we should be able to assert that this report is empty to ensure we maintain full conversion coverage of Tags properties.

Add a fallback message to properties which have no type information (represented by "Any") on where to look up additional information in the CloudFormation docs.


provider/pkg/schema/gen.go
```diff
@@ -595,7 +600,7 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 		},
 	}
 
-	return &p, &metadata, nil
+	return &p, &metadata, reports, nil
 }
 
 // context holds shared information for a single CF JSON schema.
```
thomas11, line 606: Not part of this PR but `context` is a confusing type name given the prevalence of `context.Context`

provider/pkg/schema/gen.go
```diff
@@ -936,6 +942,22 @@ func (ctx *context) propertySpec(propName, resourceTypeName string, spec *jssche
 		}
 	}
 
+	if propName == "Tags" {
```
thomas11, line 945: "Tags" is sure to always be in that casing?

provider/pkg/schema/gen.go
```diff
@@ -936,6 +942,22 @@ func (ctx *context) propertySpec(propName, resourceTypeName string, spec *jssche
...
+		switch ctx.GetTagsStyle(typeSpec, spec) {
+		case TagsStyleUntyped:
+			if propertySpec.Description != "" {
+				propertySpec.Description += "\n\n"
+			}
+			propertySpec.Description += "No additional type information is available in the schema for this property, please refer to the Cloud Control documentation for the expected schema."
+		case TagsStyleStringMap:
+			// Nothing to do
+		case TagsStyleKeyValueArray:
+			// Swap referenced type to shared definition and remove custom type.
```
thomas11, line 955: Is that a TODO?

provider/pkg/schema/gen_tags.go
```diff
@@ -0,0 +1,35 @@
...
+type TagsStyle string
+
+const (
+	TagsStyleUnknown       TagsStyle = ""
+	TagsStyleUntyped       TagsStyle = "untyped"
+	TagsStyleStringMap     TagsStyle = "stringMap"
+	TagsStyleKeyValueArray TagsStyle = "keyValueArray"
+)
+
+func (ctx *context) GetTagsStyle(typeSpec *pschema.TypeSpec, originalSpec *jsschema.Schema) TagsStyle {
```
thomas11, line 17: Is it intended that it never returns `TagsStyleKeyValueArray`? If so, a comment might be helpful.

provider/pkg/schema/gen_tags.go
```diff
@@ -0,0 +1,35 @@
...
+	if typeSpec == nil || originalSpec == nil {
+		return TagsStyleUnknown
+	}
+	// Check for "Any" ref
+	if typeSpec.Ref == "pulumi.json#/Any" {
+		return TagsStyleUntyped
+	}
+	switch typeSpec.Type {
+	case "object":
```
thomas11, line 26: [nit] Could simplify with `case "object" && typeSpec.AdditionalProperties != nil && typeSpec.AdditionalProperties.Type == "string"` and otherwise fall through to `default`. Unless you structured it that way to add more cases later.

provider/pkg/schema/gen_tags.go
thomas11, line 17: Maybe complex enough to warrant a table-based test, especially if there will be more cases later.

provider/pkg/schema/gen.go
danielrbradley: Yes I agree 😑

provider/pkg/schema/gen_tags.go
danielrbradley: Yeah, this is again a placeholder for the follow-up PR which implements more cases.

provider/pkg/schema/gen_tags.go
danielrbradley: Yeah, I ended up scrapping the switch as the type name is actually not a clean way of detecting each case.

thomas11 (approved):

provider/pkg/schema/gen.go
mjeffryes: Actually, I think technically the name of the tags field is declared individually for each resource (eg. [here](https://github.com/pulumi/pulumi-aws-native/blob/96066277e4f414aade19612d31dc58fa78e91113/aws-cloudformation-schema/aws-autoscaling-autoscalinggroup.json#L6)) There are only a handful of resources that use something other than "Tags" though: 
```
rg tagProperty aws-cloudformation-schema | rg -v '/properties/Tags' | head
aws-cloudformation-schema/aws-datapipeline-pipeline.json:    "tagProperty" : "/properties/PipelineTags"
aws-cloudformation-schema/aws-cognito-userpool.json:    "tagProperty" : "/properties/UserPoolTags"
aws-cloudformation-schema/aws-backup-backupvault.json:    "tagProperty" : "/properties/BackupVaultTags"
aws-cloudformation-schema/aws-backup-reportplan.json:    "tagProperty" : "/properties/ReportPlanTags"
aws-cloudformation-schema/aws-efs-accesspoint.json:    "tagProperty" : "/properties/AccessPointTags"
aws-cloudformation-schema/aws-efs-filesystem.json:    "tagProperty" : "/properties/FileSystemTags"
aws-cloudformation-schema/aws-backup-backupplan.json:    "tagProperty" : "/properties/BackupPlanTags"
aws-cloudformation-schema/aws-backup-framework.json:    "tagProperty" : "/properties/FrameworkTags"
```

provider/pkg/schema/gen.go
```diff
@@ -1347,3 +1369,35 @@ func GatherSemantics(schemaDir string) (SemanticsSpecDocument, error) {
 	}
 	return semanticsDocument, nil
 }
+
+type Reports struct {
```
mjeffryes, line 1373: nit. I think reports should be in it's own file, gen.go is already a bit of a monster.

provider/pkg/schema/gen.go
```diff
@@ -936,6 +942,22 @@ func (ctx *context) propertySpec(propName, resourceTypeName string, spec *jssche
...
 
+	if propName == "Tags" {
+		switch ctx.GetTagsStyle(typeSpec, spec) {
+		case TagsStyleUntyped:
+			if propertySpec.Description != "" {
+				propertySpec.Description += "\n\n"
+			}
+			propertySpec.Description += "No additional type information is available in the schema for this property, please refer to the Cloud Control documentation for the expected schema."
+		case TagsStyleStringMap:
+			// Nothing to do
```
mjeffryes, line 953: So IIUC, the idea is that these `map[string]string` shaped Tag properties will just stay `map[string]string`, but in the next review, we'll start converting things that are a list of `{ key:string, value: string }` to be `[]Tag` rather than `[]SomeResourceTag`

provider/pkg/schema/gen.go
```diff
@@ -936,6 +942,22 @@ func (ctx *context) propertySpec(propName, resourceTypeName string, spec *jssche
 		}
 	}
 
+	if propName == "Tags" {
+		switch ctx.GetTagsStyle(typeSpec, spec) {
+		case TagsStyleUntyped:
+			if propertySpec.Description != "" {
+				propertySpec.Description += "\n\n"
+			}
+			propertySpec.Description += "No additional type information is available in the schema for this property, please refer to the Cloud Control documentation for the expected schema."
```
mjeffryes, line 951: do we really need to add this? And if we think we do, why not add it to all properties of type "Any"?

provider/pkg/schema/metadata.go
```diff
@@ -20,6 +20,7 @@ type CloudAPIResource struct {
 	Required          []string                        `json:"required,omitempty"`
 	CreateOnly        []string                        `json:"createOnly,omitempty"`
 	WriteOnly         []string                        `json:"writeOnly,omitempty"`
+	UnorderedArrays   []string                        `json:"unorderedArrays,omitempty"`
```
mjeffryes, line 23: I didn't see this used in here; I assume it's logically part of the follow up review?

provider/pkg/schema/gen.go
mjeffryes: And of course just for inconsistency, there are resources with a `Tags` field that don't  include the tagging metadata block. I feel like we would be more justified in skipping those ones though.

mjeffryes (commented):
No must change concerns, but I do wonder if we should maybe use the tagging metadata rather than the field name for identifying which properties are supposed to be tags

provider/pkg/schema/metadata.go
danielrbradley: Yes, was going to be in the follow-up though I think I overlooked this in the next PR 👀

provider/pkg/schema/gen.go
danielrbradley: Nope, in the design discussion Luke pushed back strongly on unifying tag shapes and to just leave them as-is .. so we're only reducing the duplication of types now.

provider/pkg/schema/gen.go
danielrbradley: Hmm, maybe not. The first version of the message was a little more specific - indicating it was likely either a string map or a list of key-value objects, but I wasn't happy that the wording was clear enough so simplified it back to this, but perhaps we take it out.

danielrbradley: Ok, it now adds something like this for all properties:

```
Search the [CloudFormation User Guide](https://docs.aws.amazon.com/cloudformation/) for `AWS::Oam::Link` for more information about the expected schema for this property.
```

provider/pkg/schema/metadata.go
danielrbradley: Have taken this out for now. This will be for customising diff correctly when they're arrays.

Issue #1052, opened 2023-08-25
annoyance: golang resource tags all different incompatible types


ploxiln:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

In the pulumi aws-classic golang SDK, every resource New function Args struct has Tags as a `pulumi.StringMapInput` for which you can provide a `pulumi.StringMap`. Example: [rds.InstanceArgs](https://pkg.go.dev/github.com/pulumi/pulumi-aws/sdk/v5@v5.42.0/go/aws/rds#InstanceArgs). You can conveniently use `ToStringMap()` to just pass a concise `map[string]string`, or filling out the `StringMap{ ... }` inline isn't too bad either, just need to `pulumi.String()` any literal values, and you can then mix in any `pulumi.StringOutput` from other resource output attributes etc. You can use common code to manage tags maps for many different resources.

In this new pulumi aws-native golang SDK, every resource New function Args struct has its own tags type. Example: [rds.DbInstanceArgs](https://pkg.go.dev/github.com/pulumi/pulumi-aws-native/sdk/go/aws/rds#DbInstanceArgs). That takes a `DbInstanceTagArrayInput`, the only type satisfying that is a `DbInstanceTagArray`, which is an array of `DbInstanceTagInput`, the only type satisfying that is `DbInstanceTagArgs`. And these are different for every other resource! For example [networkfirewall.FirewallArgs](https://pkg.go.dev/github.com/pulumi/pulumi-aws-native/sdk/go/aws/networkfirewall#FirewallArgs) takes a different incompatible type `FirewallTagArrayInput` ... it follows the exact same pattern, but it's all incompatible types. There are 3 different separate tag type hierarchies in the `networkfirewall` sub-package.

It's quite verbose to fill out literally:

```go
    networkfirewall.FirewallTagArray{
        &networkfirewall.FirewallTagArgs{
            Key: pulumi.String("env"),
            Value: pulumi.String(env),
        },
        &networkfirewall.FirewallTagArgs{
```

[...]

PR #1348, opened 2024-02-16
Generalize tag type where possible

danielrbradley:
Stacked on https://github.com/pulumi/pulumi-aws-native/pull/1347

Fixes https://github.com/pulumi/pulumi-aws-native/issues/1052

- Identify where "tags" properties are a list of types containing only a `key` and a `value` both of type "string".
- Don't generalize where additional properties are added (only Autoscaling Group has an additional property–`PropagateAtLaunch`).
- We're not requiring `additionalProperties` to be `false` as we don't support this being true right now anyway.
- We can't use the shared type if the resource has Tags listed in `createOnlyProperties` as our implementation requires each property on the tags type to then be marked as causing replacement which then affects every other resource. This excludes around 70 resources.


lukehoban:
Does this mean it will be a breaking change for Pulumi users if an optional property is added to the tags object upstream?  But that change is not a breaking change upstream? 

Are there other similar potential breaking change concerns here?

I am still somewhat nervous that this is working against the grain of a full and faithful projection of cloud control api to Pulumi.

danielrbradley:
@lukehoban 

> Does this mean it will be a breaking change for Pulumi users if an optional property is added to the tags object upstream? But that change is not a breaking change upstream?

Theoretically yes. In practice there's only a single instance of a resource having extra properties added.

In practice the upstream specs regularly introduce actual breaking changes (for us and sometimes even them) because there's no versioning of the CloudControl API so this is already something we have to navigate by regularly documenting breaking changes in the release notes for minor versions.

> I am still somewhat nervous that this is working against the grain of a full and faithful projection of cloud control api to Pulumi.

This is still a faithful 1:1 projection, but with a lot less duplication which is what users have been complaining at. Practically, IMO this is more user friendly even with the slightly increased risk of isolated breaking changes.

provider/pkg/schema/gen_tags.go
```diff
@@ -0,0 +1,73 @@
...
+
+	typeToken, hasPrefix := strings.CutPrefix(typeSpec.Items.Ref, "#/types/")
+	if !hasPrefix {
+		return false
+	}
+
+	// We can't include tags which are create-only properties because this has to be added on the tags type but this is a shared type
+	if createOnlyProps := readPropSdkNames(ctx.resourceSpec, "createOnlyProperties"); createOnlyProps.Has("tags") {
+		return false
+	}
```
mjeffryes, line 56: I really don't like how this depends on the field name being "tags"

provider/pkg/schema/gen_tags.go
```diff
@@ -0,0 +1,73 @@
...
+	if typeSpec == nil || typeSpec.Items == nil || typeSpec.Items.Ref == "" {
+		return false
+	}
+
+	typeToken, hasPrefix := strings.CutPrefix(typeSpec.Items.Ref, "#/types/")
+	if !hasPrefix {
+		return false
+	}
+
+	// We can't include tags which are create-only properties because this has to be added on the tags type but this is a shared type
```
mjeffryes, line 53: Frustrating. The way we represent create-only properties should be a property of the resource, not a type annotation!

danielrbradley: Agreed. It's even more annoying because when the SDKs get rendered they _are_ then just a set of properties on the resource.

provider/pkg/schema/gen_tags.go
danielrbradley: I think this is safe enough given that we're only calling into this method when we're already looking at a property named "tags". We don't want to expand this to other property names because the type is specially named "Tag" so other types with the same fields wouldn't use this logic.

provider/pkg/schema/gen_tags.go
danielrbradley: Here's an existing issue to add a better option for this in the schema: 
- https://github.com/pulumi/pulumi/issues/9158

Our other option would be to ditch using the engine's diff and instead implement this in the provider so we can customise when we do replacements. But that's quite a large piece to take on.

provider/pkg/schema/gen_tags.go
danielrbradley: Ok, fixed ... needed this now we changed the logic to also lookup alternative tags fields from the schema.

provider/pkg/schema/gen_tags.go
```diff
@@ -44,14 +45,41 @@ func (ctx *context) GetTagsStyle(typeSpec *pschema.TypeSpec, originalSpec *jssch
...
+	typeToken, hasPrefix := strings.CutPrefix(typeSpec.Items.Ref, "#/types/")
+	if !hasPrefix {
+		return false
+	}
+
+	if refType, ok := ctx.pkg.Types[typeToken]; ok {
+		keyProp, keyPropExists := refType.Properties["key"]
+		valueProp, valuePropExists := refType.Properties["value"]
+		// Check if the type has exactly two properties, "key" and "value", both of type "string"
+		for k := range refType.Properties {
```
thomas11, line 89: [nit] This loop could just be `keyPropExists && valuePropExists && len(refType.Properties) == 2`, right?

provider/pkg/schema/gen_tags.go
thomas11: What if we had `aws-native:index:Tag` and `aws-native:index:CreateOnlyTag`?

thomas11 (approved):

provider/pkg/schema/gen_tags.go
danielrbradley: Great minds ... that's exactly what I ended up adding for now!

[Implement shared type for all tags that cause replacement](https://github.com/pulumi/pulumi-aws-native/pull/1348/commits/e847c2c1fc6f508684441a1b591c25c8f611dcf0)

provider/pkg/schema/gen_tags.go
danielrbradley: Ah yup - much nicer 👍

PR #1355, opened 2024-02-20
ADD more descriptive explanation for missing aws native region config

synoet:
I got an error message from Pulumi saying I must set the AWS-native region value. I tried setting it with `pulumi config set aws-native:region us-east-2` multiple times, and it updated in my Pulumi.<stack>.yaml. Yet, the error persisted during `pulumi refresh`.

Setting the `AWS_REGION` environment variable resolved the problem. While this doesn't address the root cause, it offers a working alternative solution.

https://github.com/pulumi/pulumi-aws-native/issues/1356


danielrbradley (approved):
Happy with this minor wording change - tests are failing due to an issue that's already fixed on master so this should pass fine once merged.

Thanks for the contribution @synoet 🎉

PR #1363, opened 2024-02-22
Fix generating types for refs to map types

danielrbradley:
When parsing a property which has a ref to a type which is just a map (using addtionalProperties or patternProperties) we were returning an untyped "object".

This adds a special case to check if the referenced type could be represented as a map first and generated any referenced value types.

Disable generating nested collection types (e.g. `Map<Map<Array<string>>>`) as this currently breaks Go codegen. See:
- https://github.com/pulumi/pulumi/issues/15478

This fixes 143 resource properties.

Follow-up to:
- https://github.com/pulumi/pulumi-aws-native/pull/1348


provider/pkg/schema/gen.go
```diff
@@ -1043,6 +1043,35 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
...
 			return nil, errors.Errorf("definition %s not found in resource %s", schemaName, ctx.cfTypeName)
 		}
+		var mapType *jsschema.Schema
+		if typeSchema.AdditionalProperties != nil && typeSchema.AdditionalProperties.Schema != nil {
+			mapType = typeSchema.AdditionalProperties.Schema
+		} else if typeSchema.PatternProperties != nil && len(typeSchema.PatternProperties) == 1 {
+			// TODO: Capture pattern add add to documentation or even provider metadata for early validation feedback.
+			// https://github.com/pulumi/pulumi-aws-native/issues/1346
+			for _, schema := range typeSchema.PatternProperties {
+				mapType = schema
```
thomas11, line 1053: This is just the last of the patterns? Not sure I understand what's happening here.

danielrbradley: We're only looking at this if there's only a single pattern property in the map so we're just accessing the single value in the map here. It's not a list so I can't just do `typeSchema.PatternProperties[0]`.

thomas11 (approved):

PR #1367, opened 2024-02-23
Identify all tag styles

danielrbradley:
1. Add detection of last two observed variations.
2. Add test to fail unit test if new unexpected tag styles are found.

The purpose of complete coverage of all tags styles is to maintain quality of tag accessibility. Additionally, this will allow us to reliably apply DefaultTags in a future PR (see https://github.com/pulumi/pulumi-aws-native/issues/107).

### DIFF
--- a/provider/pkg/schema/gen_tags.go
+++ b/provider/pkg/schema/gen_tags.go
@@ -62,56 +62,53 @@ func (ctx *context) GetTagsStyle(propName string, typeSpec *pschema.TypeSpec) Ta
 		return TagsStyleStringMap
 	}
 
-	if ctx.tagStyleIsKeyValueArray(propName, typeSpec, false /* includeExtraProperties */, false /* allowOneOf */) {
-		if ctx.isPropCreateOnly(propName) {
+	if isKeyValueArray, style := ctx.tagStyleIsKeyValueArray(propName, typeSpec); isKeyValueArray {
+		if style == TagsStyleKeyValueArray && ctx.isPropCreateOnly(propName) {
 			return TagsStyleKeyValueCreateOnlyArray
 		}
-		return TagsStyleKeyValueArray
-	}
-
-	if ctx.tagStyleIsKeyValueArray(propName, typeSpec, true /* includeExtraProperties */, false /* allowOneOf */) {
-		return TagsStyleKeyValueArrayWithExtraProperties
-	}
-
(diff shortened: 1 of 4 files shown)


provider/pkg/schema/gen_tags.go
```diff
@@ -56,40 +62,73 @@ func (ctx *context) GetTagsStyle(propName string, typeSpec *pschema.TypeSpec) Ta
 		return TagsStyleStringMap
 	}
 
-	if ctx.tagStyleIsKeyValueArray(propName, typeSpec, false /* includeCreateOnly */) {
+	if ctx.tagStyleIsKeyValueArray(propName, typeSpec, false /* includeExtraProperties */, false /* allowOneOf */) {
```
thomas11, line 65: Calling this method three times with different combinations of boolean parameters is a bit complicated. Since the boolean params are only used to distinguish different kinds of tag arrays, could the method just return the right kind?
```go
tagArrayKind, isTagArray := ctx.tagStyleIsKeyValueArray(propName, typeSpec)
if isTagArray {
    return tagArrayKind
}
```

provider/pkg/schema/gen_tags_test.go
```diff
@@ -101,4 +105,56 @@ func TestGetTagsStyle(t *testing.T) {
...
+}
+
+func TestNoUnexpectedTagsShapes(t *testing.T) {
+	path := path.Join("..", "..", "..", "reports", "unexpectedTagsShapes.json")
+	bytes, err := os.ReadFile(path)
+	require.NoError(t, err)
+	var unexpectedTagsShapes map[string]interface{}
+	err = json.Unmarshal(bytes, &unexpectedTagsShapes)
+	require.NoError(t, err)
+	assert.Empty(t, unexpectedTagsShapes, "reports/unexpectedTagsShapes.json should be empty, which means that we need to update GetTagsStyle in gen_tags.go to cover new tags variations.")
```
thomas11, line 159: If we want to fail hard at build time, why not just directly in GetTagsStyle? So we have the complete report to look at?

thomas11 (approved):

provider/pkg/schema/gen_tags.go
danielrbradley: Yup, good thought. Will refactor a little further 👍

PR #1369, opened 2024-02-23
Default tags

danielrbradley:
Fixes https://github.com/pulumi/pulumi-aws-native/issues/107

Implement the `defaultTags` provider configuration option. When setting tags on a resource, match the tags style as defined in the metadata.


provider/pkg/schema/gen.go
```diff
@@ -851,6 +865,16 @@ func (ctx *context) gatherResourceType() error {
 		if err != nil {
 			return err
 		}
+		if hasTags && prop == tagsProp {
```
thomas11, line 868: This added block seems to be identical to the one above? Can we extract it?

provider/pkg/schema/gen_tags.go
```diff
@@ -29,24 +31,62 @@ const (
...
-	}
-	asInterface, ok := tagging.(map[string]interface{})
-	if !ok {
-		return "", false
+func (ts TagsStyle) IsStringMap() bool {
+	return ts == TagsStyleStringMap
+}
+
+func (ts TagsStyle) IsKeyValueArray() bool {
+	return strings.HasPrefix(string(ts), "keyValueArray")
```
thomas11, line 39: Would be a tad safer to use a `const keyValueArrayPrefix` here and in the names

provider/pkg/schema/gen_tags.go
```diff
@@ -29,24 +31,62 @@ const (
...
+	default: // Unknown
+		ctx.reports.UnexpectedTagsShapes[ctx.resourceToken] = spec
 	}
-	tagProperty, ok := asInterface["tagProperty"]
-	if !ok {
-		return "", false
+}
+
+func GetTagsProperty(originalSpec *jsschema.Schema) (string, bool) {
+	fromTaggingMetadata, ok := func() (string, bool) {
```
thomas11, line 63: What's the reason for using an inline function that's instantly run? Isn't that the same as inlining the function body?

provider/pkg/provider/provider.go
```diff
@@ -459,6 +457,17 @@ func (p *cfnProvider) Configure(ctx context.Context, req *pulumirpc.ConfigureReq
 		}
 	}
 
+	if defaultTagsJson, ok := vars["aws-native:config:defaultTags"]; ok {
+		var defaultTags map[string]string
+		err := json.Unmarshal([]byte(defaultTagsJson), &defaultTags)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal 'forbiddenAccountIds' config: %w", err)
```
thomas11, line 464: wrong copied name

provider/pkg/provider/provider.go
```diff
@@ -459,6 +457,17 @@ func (p *cfnProvider) Configure(ctx context.Context, req *pulumirpc.ConfigureReq
...
+		}
+		p.defaultTags = defaultTags
+	} else {
+		p.defaultTags = make(map[string]string)
```
thomas11, line 468: nil maps are usable for lookups, do we need an actual map?

provider/pkg/provider/provider_defaultTags_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"testing"
+
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDefaultTags(t *testing.T) {
```
thomas11, line 11: There are some more cases that would be good to cover IMO, like not overwriting existing tags with default tags.

provider/pkg/provider/provider.go
danielrbradley: Sure, I'll change to a nil assignment. It's kinda an edge case, but when debugging the provider, if we don't allocate a nil here it causes a weird bug when we remove the default tags from the config as the old value will then remain set if we don't explicitly unset it.

provider/pkg/schema/gen.go
danielrbradley: Done 👍

thomas11 (approved):

fitz-vivodyne:
Is this supported now?
The [documentation](https://www.pulumi.com/registry/packages/aws-native/installation-configuration/#configuration-options) still states that it isn't

danielrbradley:
Yes, this works in v0.100.0

Issue #1191, opened 2023-11-27
Environment variables take precedence over explicit configuration


midu-git:
### What happened?

When using the `aws-native` provider setting `profile` explicitly while having the environment variables `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN` set, the provider ignores the profile configuration and uses the environment variables instead.

### Example

Get a session and set environment, e.g. (zsh)
```shell
STS=(`aws sts assume-role --role-arn 'SOME_ROLE_ARN' --role-session-name 'test' --query '[Credentials.AccessKeyId,Credentials.SecretAccessKey,Credentials.SessionToken]' --duration-seconds ${2:-3600} --output text`)
export AWS_ACCESS_KEY_ID="${STS[1]}"
export AWS_SECRET_ACCESS_KEY="${STS[2]}"
export AWS_SESSION_TOKEN="${STS[3]}"
```
Create your `aws-native` provider using
```typescript
function buildAwsNativeProvider(
    profile: string,
    accountId: string,
    region: awsNative.Region
): awsNative.Provider {
    return new awsNative.Provider(`aws-native-${accountId}-${region}`, {
        profile: profile,
        allowedAccountIds: [accountId],
        region: region,
    });
}
```
Alternatively, set a default provider in `Pulumi.stackname.yaml`
```yaml
config:
  aws:allowedAccountIds:
    - "YOUR_ACCOUNT_ID"
  aws:profile: YOUR_PROFILE
  aws-native:allowedAccountIds:
    - "YOUR_ACCOUNT_ID"
  aws-native:profile: YOUR_PROFILE
```
Creating a resource in this setup results in using the identity provided by the environment variables instead of the one defined by `profile`. It is worth mentioning, that the `aws` provider correctly uses `profile`.

### Output of `pulumi about`

CLI
Version      3.94.2
Go Version   go1.21.4
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host
OS       darwin
Version  14.1.1
Arch     arm64

This project is written in nodejs: executable='/opt/homebrew/bin/node' version='v18.18.2'

Tested with `aws-native` plugins:
```
aws-native  resource  0.86.0
aws-native  resource  0.81.0
```
and `aws` plugins:
```
aws         resource  6.10.0
aws         resource  6.7.0
aws         resource  6.5.0
```

### Contributing

[...]

PR #1378, opened 2024-02-28
Fix precedence of credential sources

danielrbradley:
When both environment variables such as `AWS_ACCESS_KEY_ID` are present, and a profile is set explicitly in provider configuration, we will now choose the explict provider configuration.

Don't load AWS environment variables ourselves as this is implemented in AWS's LoadDefaultConfig method already where it has the correct preference to use the named profile over any environment variables. This precidence is defined here: https://github.com/aws/aws-sdk-go-v2/blob/58cf6509525a12d64fd826da883bfdbacbd2f00e/config/resolve_credentials.go#L102-L134

When we were parsing the access key environment variables ourselves, it appeared to AWS's library that these were not just in the environment, but specified manually by us alongside the profile. When the profile is defined alongside an explicit access key, the profile is ignored. However, if only the profile is specified by the user, but access keys are available ambiently via the environment, the profile will be used instead.

We don't currently have any good facility to test the various difference configuration variations so have tested this manually by altering local configuration.

We might also be able to remove the custom checking for AWS_REGION, AWS_DEFAULT_REGION and AWS_SHARED_CREDENTIALS_FILE for the same reason, but this will require further manual testing.

Fixes https://github.com/pulumi/pulumi-aws-native/issues/1191


mjeffryes (approved):
This looks like the correct behavior to me, but I it is a backwards incompatible change, so we should probably at least make a note of it in the CHANGELOG

provider/pkg/provider/provider.go
```diff
@@ -316,13 +316,13 @@ func (p *cfnProvider) Configure(ctx context.Context, req *pulumirpc.ConfigureReq
 
 	var accessKey, secretKey, token string
 
-	if v, ok := varsOrEnv(vars, "aws-native:config:accessKey", "AWS_ACCESS_KEY_ID"); ok {
```
thomas11, line 319: This would be a good place for a comment on the `AWS_` vars and that the SDK is already checking them. For future devs debugging issues.

thomas11 (approved):

provider/pkg/provider/provider.go
danielrbradley: Done ✅

PR #1379, opened 2024-02-29
Prepare for v0.98.0 release

rquitales:
Schema changes identified in #1376


thomas11 (approved):
Due to a flaw in this repo's CI setup, this PR will not run tests because CHANGELOG.md is excluded from requiring them, but without tests you cannot merge. One workaround would be to make a minor change to the README.

Issue #655, opened 2022-10-12
Role expects RolePolicy to output string


jetersen:
### What happened?

Seems like the output from a Role generates warnings 

```
  aws-native:iam:Role (github-actions-packer-role):
    warning: Expected System.String but got System.Collections.Immutable.ImmutableDictionary`2[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Object, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]] deserializing Pulumi.AwsNative.IAM.Outputs.RolePolicy(policyDocument)
```

### Steps to reproduce

```
return await Deployment.RunAsync(() =>
{
    
    new Role("github-actions-packer-role", new RoleArgs
    {
        RoleName = "github-actions-packer-role",
        Description = "Allows GitHub Actions inside to create AMIs",
        Policies = new List<RolePolicyArgs>
        {
            new()
            {
                PolicyName = "allow-packer-to-build-ami-policy",
                PolicyDocument = new PolicyDocument
                {
                    Statement = new List<Statement>
                    {
                        new()
                        {
                            Action = new List<string>
                            {
                                "ec2:AttachVolume",
                                "ec2:AuthorizeSecurityGroupIngress",
                                "ec2:CopyImage",
                                "ec2:CreateImage",
                                "ec2:CreateKeypair",
                                "ec2:CreateSecurityGroup",
                                "ec2:CreateSnapshot",
                                "ec2:CreateTags",
                                "ec2:CreateVolume",
                                "ec2:DeleteKeyPair",
                                "ec2:DeleteSecurityGroup",
                                "ec2:DeleteSnapshot",
                                "ec2:DeleteVolume",
                                "ec2:DeregisterImage",
                                "ec2:DescribeImageAttribute",
                                "ec2:DescribeImages",
                                "ec2:DescribeInstances",
                                "ec2:DescribeInstanceStatus",
                                "ec2:DescribeRegions",
                                "ec2:DescribeSecurityGroups",
                                "ec2:DescribeSnapshots",
                                "ec2:DescribeSubnets",
                                "ec2:DescribeTags",
                                "ec2:DescribeVolumes",
                                "ec2:DetachVolume",
                                "ec2:GetPasswordData",
                                "ec2:ModifyImageAttribute",
                                "ec2:ModifyInstanceAttribute",
[...]

Issue #194, opened 2021-11-01
AWS IAM Role Creation Fails when Inline Policies are Included


phillipedwards:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

IAM Role creation fails when role definition includes inline-policies in the `policies` property.

### Steps to reproduce

Versions Used:
- Pulumi: v3.16.0
- Pulumi AWS Native: 0.2.0

1. Create a new Pulumi AWS-Native Python project
2. Using below code:
    ```
	  from pulumi_aws_native import iam
	  import json
	  
	  role = iam.Role(
	      "test-role",
	      assume_role_policy_document=json.dumps(
	      {
	          "Version": "2012-10-17",
	          "Statement": [
	              {
	                  "Action": "sts:AssumeRole",
	                  "Effect": "Allow",
	                  "Sid": "",
	                  "Principal": {"Service": ["lambda.amazonaws.com"]},
	              }
	          ],
	      }
	  ),
	      policies=[
	          iam.RolePolicyArgs(
	              policy_name="test-policy",
	              policy_document=json.dumps(
	                  {
	                      "Version": "2012-10-17",
	                      "Statement": [
	                          {
	                              "Action": ["ec2:Describe*"],
	                              "Effect": "Allow",
	                              "Resource": "*",
	                          }
	                      ],
	                  }
	              ),
	          )
	      ],
	  )
	```
3. Execute a `pulumi up`
4. See error output.

Expected: New IAM role with inline-policy is created and Pulumi program does not error.
Actual: Role is created with inline-policy is created, but error occurs and role no longer becomes manageable from Pulumi.

## Error Output
![image](https://user-images.githubusercontent.com/25461821/139752762-693637e6-7234-4c69-9090-a05b3ed7ca45.png)

Furthermore, role does exist in AWS with inline-policy attached, but any attempt to delete role fails with:
![image](https://user-images.githubusercontent.com/25461821/139752849-2c65cfa4-e3b3-41e8-961b-938af1e37539.png)

emiliza:
Linking to https://github.com/pulumi/pulumi-aws-native/issues/196 which is about the second half of the error here where delete roles fail for inline policies

mikhailshilkov:
[...]

PR #1383, opened 2024-03-01
Fix multi-type resolutions

danielrbradley:
Fixes https://github.com/pulumi/pulumi-aws-native/issues/655
Fixes https://github.com/pulumi/pulumi-aws-native/issues/194

## Fix parsing non-explicitly typed arrays

This changes 6 properties from Any to Array where there was `Items` defined in the schema, but the type was not explicitly set to `array`.

This also changes one instance of a type from `string` to `Array<string>` as the type was set to "string" but `Items` was also defined, so we now prefer the array type.

## Fix parsing of multiple types

In JSSchema the "type" field can either be a string or a list of strings (e.g. "string", "numnber", "object").
Types can also be implied from the other fields set of a schema (e.g. having "Items" indicates it could be an array).

Add handling of where we have multiple types rather than just taking the first type from the array. Model multiple types as a oneOf in the Pulumi schema.

To make a schema easier to work with, we can ensure that all types are explicitly listed in the "type" array via the `NormaliseTypes` function.

Where one of the multiple types is "object", we simplify the type to just be "Any" as this also covers any other primative too.

## SDK Change summary

- `aws-native:apigateway:Method` `requestParameters` additional properties type changed from `boolean` to `Union<boolean, string>`
- `aws-native:apigatewayv2:Route` `requestParameters` type changed from `Any` to `Array<{ required: boolean }>`
- Properties changed from `string` to `Any` (becuase it can be `Union<string, object>`:
  - `aws-native:configuration:ConfigRule` `inputParameters`
  - `aws-native:ec2:VpcEndpoint` `policyDocument`
  - `aws-native:events:Rule` `eventPattern`
- `aws-native:logs:DeliveryDestination` `deliveryDestinationPolicy` type changed from `Any` to `Array<{ deliveryDestinationName: string, deliveryDestinationPolicy: string }>`


examples/examples_dotnet_test.go
```diff
@@ -20,7 +25,31 @@ func TestSimpleCs(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestRoleCs(t *testing.T) {
+	test := getCsharpBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir:         filepath.Join(getCwd(t), "role-cs"),
+			SkipRefresh: true,
```
mjeffryes, line 32: I assume this is because refreshing turns the string policy into an object? Maybe worth a comment here about the behavior

provider/pkg/schema/gen.go
```diff
@@ -1210,37 +1212,70 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
...
+		elementType, err := ctx.propertyTypeSpec(parentName+"Item", propSchema.Items.Schemas[0])
+		if err != nil {
+			return nil, err
+		}
+		return &pschema.TypeSpec{
+			Type:  "array",
+			Items: elementType,
+		}, nil
+	}
+
```
mjeffryes, line 1236: I don't understand why you have both of these? it looks like the later is a strict superset of the former

provider/pkg/schema/gen.go
```diff
@@ -1210,37 +1212,70 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
...
+	if propSchema.Items != nil {
```
mjeffryes, line 1226: just checking for `Items` seems like it's throwing away the work you did in `NormalizeTypes` above. And we'd mistype the following schema: 
``` json
{
   "type": ["string", "array"]
   "items": { "type": "string" }
} 
```
as `[]string` when it should evaluate to `Union<string, []string>`

And in theory, we also have this problem with types that have an `object` in the type list too.

Instead, I think this logic needs to be nested in the evaluation of the list of types.

provider/pkg/schema/gen.go
```diff
@@ -1210,37 +1212,70 @@ func (ctx *context) propertyTypeSpec(parentName string, propSchema *jsschema.Sch
...
-				Type:  "array",
-				Items: elementType,
-			}, nil
-		default:
+	if len(propSchema.Type) == 1 {
+		typeSpec := parseJsonType(propSchema.Type[0])
+		return &typeSpec, nil
+	}
+
+	if len(propSchema.Type) > 1 {
```
mjeffryes, line 1243: I think, to avoid the problem with `type: ["string", "array"]` mentioned above, this check should actually come right after `NormalizeTypes` and we should recurse on `propertySpec` with a schema that's simplified to have a single element for type.

mjeffryes (commented):
Nice work cleaning these cases up. JSSchema is a weird spec. 

I noticed that there are still a few cases were schemas with multiple types could break us. AFAIK, there's not any live examples of these in the spec, so I'd be fine with punting them for now (maybe with a comment and a test case making that decision explicit).

I would however really like to see some test cases added to the gen_examples_test.go to cover all these new schema shapes we're adding support for.

examples/examples_dotnet_test.go
danielrbradley: This was not intentional, it was copied from the test above. I'll remove all of these as this PR should fix all refreshes 🤞

provider/pkg/schema/gen.go
danielrbradley: Yup, think this was a left-over from a refactoring that got reverted. Removed.

provider/pkg/schema/gen.go
danielrbradley: Yes, I think you're right that this would be a better design. I've tried to keep the change fairly conservative so far as all changes seem to have fairly wide-reaching changes. Will have a spike on if this is possible to do well. It gets a little more tricky when we can also have OneOfs too .. but perhaps we could safely assume that it can't be multi-typed if it's a one-of.

provider/pkg/schema/gen.go
danielrbradley: Yes, this does generalise `Union<string, []string>` to `[]string` currently which isn't quite as accurate technically, but is probably simpler and therefore more usable in some languages (e.g. Go) where unions are not as well supported.

thomas11: Now that we have `NormalizeTypes`, why not use it for `Items.Schemas` as well? Wouldn't fix the issues Matt pointed out?

provider/pkg/schema/jsschema_util.go
```diff
@@ -34,7 +34,22 @@ func FlattenJSSchema(sch *jsschema.Schema) jsschema.SchemaList {
 		schemas[i] = merged
 	}
 	return schemas
+}
+
+func NormaliseTypes(sch *jsschema.Schema) *jsschema.Schema {
```
thomas11, line 39: Seems like a good candidate for unit tests.

provider/pkg/schema/gen.go
danielrbradley: Alas, I've tried moving this up but this check is a less accurate version of the object and array checks meaning that we loose a lot of fidelity. I think this whole area of code needs a big rethink to address these fragile bits.

mjeffryes: Ok, thanks for giving it a look anyway. I'm fine with just leaving a comment and some tests cases that acknowledge the limitations.

provider/pkg/schema/jsschema_util.go
danielrbradley: I've gone with just adding more tests at the overall property type generation level instead.

mjeffryes (approved):

thomas11 (approved):

PR #1384, opened 2024-03-02
Remove explicit nodejs dependencies list

mjeffryes:
Cleaning up some likely copy-pasta:

While resolving a bug caused by explicit dependencies on old nodejs packages in pulumi-kubernetes (https://github.com/pulumi/pulumi-kubernetes/issues/2857), we noticed that aws-native also has these dependencies, even though it does not need them because it does not implement any overlay resources.

### DIFF
--- a/provider/pkg/schema/gen.go
+++ b/provider/pkg/schema/gen.go
@@ -419,23 +419,7 @@ func GatherPackage(supportedResourceTypes []string, jsonSchemas []jsschema.Schem
 			"github.com/pulumi/pulumi-aws-native/sdk/go/aws/aws-native": "aws",
 		},
 	})
-	p.Language["nodejs"] = rawMessage(map[string]interface{}{
-		"dependencies": map[string]string{
-			"@pulumi/pulumi":    "^3.0.0",
-			"shell-quote":       "^1.6.1",
-			"tmp":               "^0.0.33",
-			"@types/tmp":        "^0.0.33",
-			"glob":              "^7.1.2",
-			"@types/glob":       "^5.0.35",
-			"node-fetch":        "^2.3.0",
-			"@types/node-fetch": "^2.1.4",
-		},
-		"devDependencies": map[string]string{
-			"mocha":              "^5.2.0",
-			"@types/mocha":       "^5.2.5",
-			"@types/shell-quote": "^1.6.0",
-		},
-	})
+	p.Language["nodejs"] = rawMessage(map[string]interface{}{})
 	p.Language["python"] = rawMessage(map[string]interface{}{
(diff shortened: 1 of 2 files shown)


t0yv0:
I'm a little confused about what this does.

We seem to have a good place for documenting this part of schema but the current documentation is laconic and leaves me wondering.
https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#language-specific-extensions

Do these settings inform the SDK generator for the particular language, and that only? How do they affect end-users as consumers of the said SDK? 

In particular, won't removing these dependencies break the build of the Node SDK? Doesn't seem to, in the PR.

mjeffryes:
> I'm a little confused about what this does.
> 
> We seem to have a good place for documenting this part of schema but the current documentation is laconic and leaves me wondering. https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#language-specific-extensions
> 
> Do these settings inform the SDK generator for the particular language, and that only? How do they affect end-users as consumers of the said SDK?
> 
> In particular, won't removing these dependencies break the build of the Node SDK? Doesn't seem to, in the PR.

As I understand it: 
- Yes, these settings are specific to the generator for each language.
- The impact of this specific change is that we no longer add a bunch of explicit dependencies to the package.json for the node SDK (seen here in the diff: https://github.com/pulumi/pulumi-aws-native/pull/1384/files#diff-0d53ce0d334b84822a2dbdf673d0456efbacb59a76f72aabbfb1b7f8203a58f7)
- Other providers, like Kubernetes, that add overlay resources to the SDK add explicit dependencies via this mechanism so the overlays in the SDK will compile.
- AWS Native has never had any overlay resources, and thus is just fine with the default generated package.json for its Node SDK
- It's very likely, these dependencies were just copy pasted from Kubernetes when this provider was created.

r2el:
We are running into an issue updating `pulumi-aws-native` in our package.json as result we are getting compilation errors. We believe this PR may fix our issue due to the old version of `glob` is bringing in an old version of `minimatch` that is producing the error for us. 

```
node_modules/minimatch/dist/cjs/index"' has no exported member 'IOptions'.

26     interface IOptions extends minimatch.IOptions {
```
While we were searching Issues for this repo nothing seemed to come up, but when checking Pull Requests we noticed the removal of the old `glob` dependency in the files changed for this PR as a likely candidate.

t0yv0 (approved):
So we expect that removing this customization from nodejs enables the Node SDK generator to substitute some default dependencies, like "@pulumi/pulumi" that the generated SDK still depends on, but this change removes extraneous dependencies like "glob". This seems to make sense 🚢

PR #1393, opened 2024-03-07
Fix parsing provider config defaultTags

danielrbradley:
Fix the provider to parse the format as defined by the schema.

Keep a fallback in place to allow it parsing the default tags as a simple map for backward compatibility.


sdk/go/aws/provider.go
```diff
@@ -86,7 +86,7 @@ type providerArgs struct {
 	// Configuration for retrieving temporary credentials from the STS service.
 	AssumeRole *ProviderAssumeRole `pulumi:"assumeRole"`
 	// Configuration block with resource tag settings to apply across all resources handled by this provider. This is designed to replace redundant per-resource `tags` configurations. Provider tags can be overridden with new values, but not excluded from specific resources. To override provider tag values, use the `tags` argument within a resource to configure new tag values for matching keys.
-	DefaultTags *ProviderDefaultTags `pulumi:"defaultTags"`
+	DefaultTags map[string]string `pulumi:"defaultTags"`
```
t0yv0, line 89: Breaking change?

t0yv0:
So this looks good provided you do want to take a breaking change here. One reason to think twice, I checked with TF they also have this structure:

```
			"default_tags": {
				Type:        schema.TypeList,
				Optional:    true,
				MaxItems:    1,
				Description: "Configuration block with settings to default resource tags across all resources.",
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"tags": {
							Type:        schema.TypeMap,
							Optional:    true,
							Elem:        &schema.Schema{Type: schema.TypeString},
							Description: "Resource tags to default across all resources",
						},
					},
				},
			},

```

I'm not sure but I think it's because there is some code around called TagIgnoreConfig and so forth, and possibly there were some plans to make the defaultTags do more than just add tags in the TF Provider but also have some sort of "policy" impact on transforming existing tags. Perhaps that's why this was modeled as a struct so new fields could have been added later. But they never were.

Perhaps in Pulumi the answer to this sort of thing is using transformations and it's never going to be needed anyway.

t0yv0 (approved):

sdk/go/aws/provider.go
danielrbradley: Before last week this wasn't implemented. However, when doing the implementation it was done as a string map having not looked carefully enough at this schema. So since that went out, setting the properties via an explicit provider would have failed. So it's not going to break anyone.

However, it is an open question on if it would be better to change the provider behaviour to match this schema which would technically be a break for anyone who upgraded last week and set default tags via config.

t0yv0: Ah very nice. Well perhaps there's some way to do a Configure/CheckConfig/etc hook to normalize to the desired form so it's certainly not breaking anyone.

danielrbradley: I have reverted the schema change and added a backward compatible parsing fix in the provider instead.

Issue #1243, opened 2023-12-28
EKS Addon fails on update if resolve_conflicts must be OVERWRITE


tgillitzer:
### What happened?

Updating EKS Addon (coredns).  This cluster was created long before we started using Pulumi.   Used Pulumi to migrate to the EKS coredns Addon.  When this was done, I had to use the OVERWRITE option on resource_conflicts.

Now I am updating the coredns EKS Addon to a new version.   I still have the OVERWRITE option set on resource_conflicts.

The update fails, with an error of:
```Conflicts found when trying to apply. Will not continue due to resolve conflicts mode.```

I turned on debug logging in Pulumi and looked through the log.   I see that Pulumi is calling the cloud control API and doing a PATCH.   The patch body is here:
```{"ClientToken":"<removed>","Identifier":"development|coredns","PatchDocument":"[{\"op\":\"replace\",\"path\":\"/AddonVersion\",\"value\":\"v1.9.3-eksbuild.10\"}]","TypeName":"AWS::EKS::Addon"}```

When I do the same operation via the AWS console, it does a POST against /clusters/development/addons/coredns/update and include the ResourceConflicts parameter.

I tried to use the aws CLI and the clouldcontrol command to update the Addon via a PATCH like Pulumi is doing and it doesn't accept ResourceConflicts as a path that is allowed to be update.   When I do a cloudcontrol get-resource on it, it doesn't show the  ResourceConflicts parameter either.

Maybe this is an AWS Cloud Control problem and not a pulumi problem?   Does it make sense for aws-native to use the REST API instead of the CloudFormation API for this?

Thanks

### Example

Create a cluster and then add CoreDNS as an addon.   My conflict comes from the coredns config map that was created back when the cluster was originally created.   If you make a cluster today with the AWS Console, it forces you to deploy the coredns addon and that doesn't result in this same conflict.

```eks.Addon(
        resource_name="coredns-addon",
        addon_name="coredns",
        addon_version="v1.9.3-eksbuild.9",
        cluster_name=cluster.name,
        resolve_conflicts=eks.AddonResolveConflicts.OVERWRITE,
        opts=ResourceOptions(provider=aws_native_provider),
    )
```

Once the previous update is complete, change the code as below to update the addon to a new version.

```eks.Addon(
        resource_name="coredns-addon",
        addon_name="coredns",
        addon_version="v1.9.3-eksbuild.10",
        cluster_name=cluster.name,
[...]

PR #1395, opened 2024-03-07
Send unchanged write-only properties as adds on update

danielrbradley:
Write-only properties can't even be read internally within the CloudControl service so they must be included in PATCH requests to ensure the updated model validates.
- Refactor where we unmarshal inputs to avoid duplicate work.

Fixes https://github.com/pulumi/pulumi-aws-native/issues/906

Might also fix https://github.com/pulumi/pulumi-aws-native/issues/1243 but there's no code to reproduce the issue.

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -44,6 +44,21 @@ func TestUpdate(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestLambdaUpdate(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: filepath.Join(getCwd(t), "lambda-update", "step1"),
+			EditDirs: []integration.EditDir{
+				{
+					Dir:      filepath.Join(getCwd(t), "lambda-update", "step2"),
+					Additive: true,
+				},
+			},
+		})
+
+	integration.ProgramTest(t, &test)
+}
+
 func TestNamingConventions(t *testing.T) {
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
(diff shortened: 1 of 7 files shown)


provider/pkg/provider/provider.go
```diff
@@ -754,7 +754,17 @@ func (p *cfnProvider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pu
 	label := fmt.Sprintf("%s.Diff(%s)", p.name, urn)
 	glog.V(9).Infof("%s executing", label)
 
-	diff, err := p.diffState(req.GetOlds(), req.GetNews(), label)
+	newInputs, err := plugin.UnmarshalProperties(req.GetNews(), plugin.MarshalOptions{
+		Label:        fmt.Sprintf("%s.properties", label),
+		KeepUnknowns: true,
+		RejectAssets: true,
```
t0yv0, line 760: Never use assets for lambda code?

examples/examples_nodejs_test.go
```diff
@@ -44,6 +44,21 @@ func TestUpdate(t *testing.T) {
...
+			Dir: filepath.Join(getCwd(t), "lambda-update", "step1"),
+			EditDirs: []integration.EditDir{
+				{
+					Dir:      filepath.Join(getCwd(t), "lambda-update", "step2"),
+					Additive: true,
+				},
+			},
+		})
+
+	integration.ProgramTest(t, &test)
```
t0yv0, line 59: Does this need some more asserts? Or it used to panic or some such before the change?

t0yv0:
A bit too much indirection and lack of familiarity for me to review this properly but I tried and added a few comments.

examples/examples_nodejs_test.go
danielrbradley: The update was a hard failure before this change.

provider/pkg/provider/provider.go
danielrbradley: Not currently - Asset support is still a WIP PR which will change these settings everywhere.

thomas11 (approved):

PR #1397, opened 2024-03-08
Handle secrets in update diffs

mikhailshilkov:


danielrbradley (commented):
Could you describe how we're now intending to handle secrets in the PR. Are we just ignoring all secretness for diffs? Would that mean that secret values will then be shown in detailed diff by accident?

I think Eron's recent work might be helpful here - where it also checks if secrets or unknowns were present while decoding. See: https://github.com/pulumi/pulumi-go-provider/tree/main/resourcex

A nit-pick on the testing side: we're trying to move away from monolithic unit tests where there's a single large input testing many scenarious and instead break these up into individual cases for better maintainability. This enables the test cases to outline why we're testing each kind of input as well as giving much more helpful indication as to what specific test has failed and why in the future.

mikhailshilkov:
@danielrbradley 

> Could you describe how we're now intending to handle secrets in the PR. Are we just ignoring all secretness for diffs? Would that mean that secret values will then be shown in detailed diff by accident?

This PR has no effect on the diff functionality:

1. The PR only changes the `Update` function and what it sends on the wire.
2. For better or worse, the Diff function currently [returns no data](https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/provider/provider.go#L785) to the engine, so it has no influence of secretness of values that the CLI displays.
3. The provider doesn't actually need to signal secretness for the engine. AWS [CC signals to accept secrets](https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/provider/provider.go#L508), but I'm not actually sure why. Azure Native does not do so, so it avoids the complication of dealing with secrets altogether.

> I think Eron's recent work might be helpful here - where it also checks if secrets or unknowns were present while decoding. 

Ah, nice, I'll see if I can use `resourcex.Decode()`.

> we're trying to move away from monolithic unit tests

Do you mean `TestDiffToPatch`? Or what's considered monolithic? Is this idea written down anywhere?

danielrbradley:
@mikhailshilkov great, thanks for the clarification. That all makes sense - and I'm happy this is a low-risk change then 👍 
 
> Do you mean `TestDiffToPatch`? Or what's considered monolithic? Is this idea written down anywhere?

Yes, mostly `TestDiffToPatch` but also probably `TestToPlainMap` too. It's tricky to know why each bit of data was specified the way it was, and if we change any of the data, if we'll accidentally loose coverage. `TestDiffToPatch` should probably be a suite of subtests from what I can see:
1. Secrets update
2. Numbers update (maybe delete too e.g. deleting a zero)
3. Bools
4. Nested objects (with each of the above)
5. Removals

It's not written down per-se, but there's some nice examples of this. Here's an example of applying the pattern on a non-trivial area where each scenario is written as its own case rather than having one large set of inputs and one large expected result:

https://github.com/pulumi/pulumi-aws-native/blob/b46b6fb7d2867a28f7347025c655c57835ac533d/provider/pkg/schema/gen_examples_test.go#L101-L109

mikhailshilkov:
@danielrbradley Made a few changes now:

1. Referenced the `resourcex` package (preview version, will update to stable once a release is out)
2. Use its `Decode`/`DecodeValue` to decode properties with secrets.
3. Kept the existing unit test intact but added a separate one for secretness decoding.

PTAL

examples/update/step2/index.ts
```diff
@@ -8,6 +9,10 @@ const logGroup = new aws.s3.Bucket("bucket", {
       key: "foo",
       value: "buzz", // <-- this value has changed
     },
+    {
+      key: "secretfoo",
+      value: pulumi.secret("secretbuzz"), // <-- this value has changed
+    },
```
danielrbradley, line 15: How do we assert that this value has been updated and not missed?

danielrbradley (approved):
Yes, think this all makes sense. Just one question on ensuring our test is definitely proving that this doesn't regress.

examples/update/step2/index.ts
mikhailshilkov: Good question, currently we don't. The failure was at the time of serialization, so we prove we don't get that error. Let me see if we have prior art on update validation.

mikhailshilkov: Done now: exported the secret value and validated it from the test.

PR #1402, opened 2024-03-11
Remove unsupported resources

danielrbradley:
Use the same schema for the SDK and docs:

- Switch to using `pulumi package gen-sdk`.
  - Remove unused language-specific code from the gen binary.
- Generate a single schema - just what we use in the docs:
  - Exclude unsupported resources.
  - Include examples in descriptions.
- Always clean SDK directory before generation to avoid left-over files.

Generate SDKs:

- Warning text changes in all files.
- Resource descriptions include examples.
- Unsupported resources are removed.
- Remove obsolete cf2pulumi full schema as we only use the uncommitted compressed schemas now.


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -73,68 +69,57 @@ func main() {
...
-	if err != nil {
-		panic(fmt.Sprintf("error gathering semantics: %v", err))
-	}
-	fullSpec, _, _, err := schema.GatherPackage(supportedTypes, jsonSchemas, true, &semanticsDocument)
-	if err != nil {
-		panic(fmt.Sprintf("error generating schema: %v", err))
-	}
-	fullSpec.Version = version
-
 	for _, language := range strings.Split(languages, ",") {
```
mikhailshilkov, line 72: Why do we still need this loop if the only remaining language is "schema"? Worth simplifying?

mikhailshilkov (approved):
LGTM, thank you

PR #1406, opened 2024-03-12
Fix extension resource

### LINKED ISSUE #1405: Test and fix ExtensionResource
- Add support for updates.
- Add integration tests to prevent regressions.

Currently panics and crashes the provider during Deletion:

```
 -  aws-native:index:ExtensionResource log-test deleting (1s) error: error reading from server: EOF
 -  aws-native:index:ExtensionResource log-test **deleting failed** error: error reading from server: EOF
    pulumi:pulumi:Stack untyped-py-p-it-daniels-ma-untyped-py-825813c3  error: update failed
    pulumi:pulumi:Stack untyped-py-p-it-daniels-ma-untyped-py-825813c3 **failed** 1 error; 22 messages
Diagnostics:
  aws-native:index:ExtensionResource (log-test):
    error: error reading from server: EOF

  pulumi:pulumi:Stack (untyped-py-p-it-daniels-ma-untyped-py-825813c3):
    panic: interface conversion: interface {} is resource.PropertyMap, not *resource.Secret
    goroutine 65 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.SecretValue(...)
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.107.0/go/common/resource/properties.go:460
    github.com/pulumi/pulumi-aws-native/provider/pkg/provider.parseCheckpointObject(...)
        /Users/dan/Dev/pulumi-aws-native/provider/pkg/provider/provider.go:1378
    github.com/pulumi/pulumi-aws-native/provider/pkg/provider.(*cfnProvider).Delete(0x14002710a20, {0x1054ac770, 0x140025ff980}, 0x14000976000)
        /Users/dan/Dev/pulumi-aws-native/provider/pkg/provider/provider.go:1171 +0x7f0
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Delete_Handler.func1({0x1054ac770, 0x140025ff980}, {0x10532d6c0?, 0x14000976000})
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.107.0/proto/go/provider_grpc.pb.go:647 +0x74
    github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1({0x1054ac770, 0x140025ff410}, {0x10532d6c0, 0x14000976000}, 0x14000e077c0, 0x140002b9770)
        /Users/dan/go/pkg/mod/github.com/grpc-ecosystem/grpc-opentracing@v0.0.0-20180507213350-8e809c8a8645/go/otgrpc/server.go:57 +0x2e8
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Delete_Handler({0x10542bbe0?, 0x14002710a20}, {0x1054ac770, 0x140025ff410}, 0x14000d94400, 0x14000d92ce0)
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.107.0/proto/go/provider_grpc.pb.go:649 +0x12c
    google.golang.org/grpc.(*Server).processUnaryRPC(0x140003c43c0, {0x1054ac770, 0x140025ff380}, {0x1054b9e20, 0x14000190d00}, 0x1400261e900, 0x14002603d40, 0x106826ca8, 0x0)
        /Users/dan/go/pkg/mod/google.golang.org/grpc@v1.59.0/server.go:1343 +0xb8c
    google.golang.org/grpc.(*Server).handleStream(0x140003c43c0, {0x1054b9e20, 0x14000190d00}, 0x1400261e900)
        /Users/dan/go/pkg/mod/google.golang.org/grpc@v1.59.0/server.go:1737 +0x990
    google.golang.org/grpc.(*Server).serveStreams.func1.1()
        /Users/dan/go/pkg/mod/google.golang.org/grpc@v1.59.0/server.go:986 +0x88
    created by google.golang.org/grpc.(*Server).serveStreams.func1 in goroutine 25
        /Users/dan/go/pkg/mod/google.golang.org/grpc@v1.59.0/server.go:997 +0x160

    error: update failed
```

danielrbradley:
Fixes #1405 

Rebuild the ExtensionResource to support update and refresh operations and allow the user to specify metadata where required for default tags and write only properties.

This also re-architects how the extension resource is integrated into the main provider by extracting the implementation for the extension resource into its own class.


provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func NewExtensionResource(host *provider.HostClient, client client.CloudControlClient) *extensionResource {
+	return &extensionResource{
+		host:   host,
+		client: client,
+	}
+}
+
+func (r *extensionResource) Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs resource.PropertyMap, err error) {
```
mikhailshilkov, line 58: I'm missing unit tests around everything in this file. With client mocking, it should be very doable, and very much warranted for the logic complexity-wise.

examples/examples_py_test.go
```diff
@@ -21,6 +21,15 @@ func TestSimplePython(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestUntypedPython(t *testing.T) {
+	test := getPythonBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: filepath.Join(getCwd(t), "untyped-py"),
+		})
+
+	integration.ProgramTest(t, &test)
```
mikhailshilkov, line 30: I suggest we extend this test to exercise Create, Update, Refresh, and Delete for the resource, so that we cover all operations. Also, test what's written in the `outputs` stack output.

provider/pkg/schema/extension_resource.go
```diff
@@ -19,6 +32,169 @@ func (e *ExtensionResource) ToSpec() CloudAPIResource {
...
+
+	id, resourceState, err := r.client.Create(ctx, extensionResource.Type, extensionResource.Properties)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create resource: %w", err)
+	}
+
+	return id, extensionResource.ToOutputs(resourceState), nil
+}
+
+func (r *extensionResource) Read(ctx context.Context, urn resource.URN, id string, oldInputs, oldState resource.PropertyMap) (resource.PropertyMap, resource.PropertyMap, bool, error) {
```
mikhailshilkov, line 73: This entire method needs better docs and tests. I'm mostly skipping its logic for now in my review.

provider/pkg/custom/interface.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+type CustomResource interface {
+	// Create creates a new resource in the cloud provider and returns its unique identifier and outputs.
+	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs resource.PropertyMap, err error)
+	// Read returns the outputs and the updated inputs of the resource.
+	Read(ctx context.Context, urn resource.URN, id string, oldInputs, oldOutputs resource.PropertyMap) (resource.PropertyMap, resource.PropertyMap, bool, error)
```
mikhailshilkov, line 13: Passing a `urn` to these methods feels odd to me. Isn't it on a wrong level?

I think you only use it once in a logging statement, I commented on its usage specifically.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+		for _, writeOnlyProp := range extensionResource.WriteOnly {
+			if _, ok := resourceState[writeOnlyProp]; !ok {
+				oldValue, ok := extensionResource.Properties[writeOnlyProp]
+				missingProps = append(missingProps, writeOnlyProp)
+				if ok {
+					resourceState[writeOnlyProp] = oldValue
+				}
+			}
+		}
+		r.host.Log(ctx, diag.Warning, urn, fmt.Sprintf("Can't refresh write-only properties: %s", strings.Join(missingProps, ", ")))
```
mikhailshilkov, line 109: I feel like we shouldn't log from this layer. If this warning is an important feature, how will you unit-test it? I would return it up and log from the provider.

mikhailshilkov, line 109: Why is it even a warning if we can't refresh a write-only property? Is it not true by definition?

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+			jsonDiffPatch = append(jsonDiffPatch, jsondiff.Operation{
+				Type:  "add",
+				Path:  writeOnlyPropName,
+				Value: newValue,
+			})
+		}
+	}
+
+	jsonPatch := make([]jsonpatch.JsonPatchOperation, 0, len(jsonDiffPatch))
+	for _, op := range jsonDiffPatch {
```
mikhailshilkov, line 167: How is the diff calculation specific to the extension resources? Don't we take exactly the same steps for "normal" resources? Why can't the provider do that and then pass the diff down to the resource?

provider/pkg/provider/provider.go
```diff
@@ -805,54 +815,47 @@ func (p *cfnProvider) Create(ctx context.Context, req *pulumirpc.CreateRequest)
...
-			payload = v
-		} else {
-			return nil, errors.New("no 'properties' property in extension resource inputs")
-		}
-	default:
-		spec, hasSpec = p.resourceMap.Resources[resourceToken]
+	if customResource, ok := p.customResources[resourceToken]; ok {
+		var outputs resource.PropertyMap
+		id, outputs, createErr = customResource.Create(ctx, urn, inputs)
+		outputsMap = resourcex.Decode(outputs)
```
mikhailshilkov, line 825: It looks like we converted a map to PropertyMap inside the resource then back to map here.

mikhailshilkov (changes requested):
I took the first pass on this PR and left a few comments.

One high-level comment is that we need a lot more tests (a better end2end test and everything covered with unit tests).

Before asking for that though, I'd love to understand the design direction you took. What is so fundamentally different with the extension resources vs. normal resources that you took an entirely different code paths for it? It seems that this leads to a lot of logic duplication: patch calculation, reading inputs from outputs, etc. are all largely copied from the main provider flow. Why can't they stay with the main flow? In the end, we are dealing with precisely the same Cloud Control API backend, and it's just the SDK/schema shape that is different. Why fixing a panic requires this entire refactoring? The price doesn't seem worth it to me yet.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+			Description: "The style of tags this resource uses. Valid values are `stringMap` or `keyValueArray`.",
+			TypeSpec: pschema.TypeSpec{
+				Type: "string",
+			},
+		},
+	}
+}
+
+func extensionResourceOutputProperties() map[string]pschema.PropertySpec {
+	properties := extensionResourceInputProperties()
```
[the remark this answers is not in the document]
danielrbradley: I was under the impression that, depending on the SDK language, some of them auto-fill in these values and some don't. However, I've just re-tested this and it looks like that's only true if the properties are defined out outputs but not then returned by the provider. So we can just completely remove these output properties.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+			Description: "Dictionary of the extension resource properties.",
+			TypeSpec: pschema.TypeSpec{
+				Type: "object",
+				AdditionalProperties: &pschema.TypeSpec{
+					Ref: "pulumi.json#/Any",
+				},
+			},
+		},
+		"writeOnly": {
+			Description: "The names of fields which are writeOnly and should always be send during patch operations.",
```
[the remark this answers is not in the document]
danielrbradley: Well, this is literally just a field in each resource's specification called "write-only properties". There's no docs I've found on it, but it's required for these resources to be updatable without erroring. It's somewhat odd, but it's required to make this practically usable for all resources.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
+package schema
```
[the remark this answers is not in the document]
danielrbradley: My aim was to have a single file with a whole resource defined in it - schema & provider implementation. This means that, because of our existing module structure, that `schema` is the only place this can currently sit to not cause import cycles. I also wanted this in the `custom` module, but that would require extracting the core metadata types into a new module at the very least. Can do this as a follow-up maybe.

provider/pkg/custom/interface.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+type CustomResource interface {
+	// Create creates a new resource in the cloud provider and returns its unique identifier and outputs.
+	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs resource.PropertyMap, err error)
```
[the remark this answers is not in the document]
danielrbradley: Wasn't sure if other resources might need access to this additional metadata - e.g. if we support preview in the future, so it seemed like a more complete abstraction. 

On this point - should we also be doing more to propegate secret-ness between inputs and outputs more generally in this provider?

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+func NewExtensionResource(host *provider.HostClient, client client.CloudControlClient) *extensionResource {
+	return &extensionResource{
+		host:   host,
+		client: client,
+	}
+}
+
+func (r *extensionResource) Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs resource.PropertyMap, err error) {
+	var extensionResource ExtensionResource
+	_, err = resourcex.Unmarshal(&extensionResource, inputs, resourcex.UnmarshalOptions{})
```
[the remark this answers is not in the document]
danielrbradley: Unmarshal is putting the values into the strongly-typed `ExtensionResource` struct - using its annotations.

provider/pkg/schema/extension_resource.go
danielrbradley: Sure, that would be nice to have at some point.

examples/examples_py_test.go
danielrbradley: All the operations are covered by the standard program test. Yes, asserting on outputs would be nice - but the test failed hard at the first step when I originally wrote it so the first milestone was just getting it to not crash during the basic operations 😅

provider/pkg/schema/extension_resource.go
danielrbradley: Agree, at some point unit tests would be nice, but this PR focuses on just moving the provider from a point of the resource being unusable to it passing a basic integration test. I would suggest we line up some follow-up issues for testing.

provider/pkg/custom/interface.go
danielrbradley: Yes, it's for logging which a custom resource might well want to do during any given operation.

provider/pkg/schema/extension_resource.go
danielrbradley: This is also already the behaviour of all other resources. We inform the user during a refresh, which properties won't have been refreshed in case they were expecting to see a change but didn't. Perhaps we could downgrade it to info.

danielrbradley: I guess, perhaps because the user has specified this metadata themselves, we could skip this message here. Was that what you were suggesting?

provider/pkg/schema/extension_resource.go
danielrbradley: No, we can't use the same diff & patch implementation because that requires us to have metadata which includes the full schema for the resource and its types - which would not be ergonomic for a user to provide.

danielrbradley: Interestingly ... this might actually be useful to port back to the general resource too as it looks like a better overall implementation, but it's not required right now as it seems to be functioning okay.

examples/untyped-py/__main__.py
```diff
@@ -0,0 +1,13 @@
...
+
+import pulumi
+from pulumi_aws_native import ExtensionResource
+
+log_group = ExtensionResource("log-test",
+    type="AWS::Logs::LogGroup",
+    properties={
+        "RetentionInDays": 90
+    }
+)
```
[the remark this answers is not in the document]
danielrbradley: Sure, can look into extending this setup a little further.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/wI2L/jsondiff"
+)
+
+type ExtensionResource struct {
+	Type         string
+	Properties   map[string]any
+	WriteOnly    []string
+	TagsProperty string
+	TagsStyle    TagsStyle
```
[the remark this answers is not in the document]
danielrbradley: If we can know about extension resources in advance, then we should just generate them as part of the SDK.

Without these properties, quite a few resources would be completely unusable in practice. These are the minimal set required to access basic functionality without breaking.

We could group them up into another type, but it didn't seem to add any value given that all other inputs are already nested within the `Properties` property - so there's no risk of clashing. I'd argue that an extra layer of nesting is just extra avoidable typing and lines of code.

provider/pkg/schema/extension_resource.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func (e *ExtensionResource) ToSpec() CloudAPIResource {
```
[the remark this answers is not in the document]
danielrbradley: Ah yes - will remove. Was an earlier attempt to just implement this using the same logic as used for all other resources (it didn't work).

provider/pkg/custom/interface.go
```diff
@@ -0,0 +1,18 @@
+package custom
```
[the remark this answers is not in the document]
danielrbradley: I wanted the ExtensionResource here, but due to the structure of our other existing modules, it would create import cycles, so I had to move it to `schema` for now. But moving the interface into schema felt more wrong so I left it here in the hope we can fix our modules in follow-up work and move it back.

provider/pkg/schema/extension_resource.go
danielrbradley: Have removed

provider/pkg/provider/provider.go
danielrbradley: Yeah, still a little confused on where we should be using more of `PropertyMap`'s features in the main provider, but this is a pretty minor overhead to ensure the interface provides the richer type system.

danielrbradley:
> I took the first pass on this PR and left a few comments.
> 
> One high-level comment is that we need a lot more tests (a better end2end test and everything covered with unit tests).
> 
> Before asking for that though, I'd love to understand the design direction you took. What is so fundamentally different with the extension resources vs. normal resources that you took an entirely different code paths for it? It seems that this leads to a lot of logic duplication: patch calculation, reading inputs from outputs, etc. are all largely copied from the main provider flow. Why can't they stay with the main flow? In the end, we are dealing with precisely the same Cloud Control API backend, and it's just the SDK/schema shape that is different. Why fixing a panic requires this entire refactoring? The price doesn't seem worth it to me yet.

@mikhailshilkov I've addressed lots of the comments inline. I originally attempted to implement this inline and re-use as much of the code as possible but this didn't go well as many areas need to account for if the properties are directly in the resource or if they're nested inside the "properties" or "outputs" container in the ExtensionResource. This led to lots of special case switches all over the code which made it fragile and complicated. The split approach is much simpler to see that changes to the extension resource won't break the main resources and vice-versa.

The approach I'd like to take on the duplication is to pull out more of these inline pieces of code into their own functions which both can use instead - but I'm leaving that as a follow-up task, secondary to getting the resource in a working state for launch.

Also, to note - this isn't just a fix of a single panic - this resource was fundamentally unusable in multiple areas and this is a complete rewrite from scratch of the whole provider implementation for it.

One outstanding item this has surfaced is the default tags support which got lost in the rebase. I'll follow up Monday with a fix for that.

provider/pkg/schema/extension_resource.go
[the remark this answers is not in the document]
mikhailshilkov: > I also wanted this in the custom module, but that would require extracting the core metadata types into a new module at the very least. 

Then we should do so. The current status quo looks pretty bad to me, I'm not comfortable proceeding with it.

provider/pkg/custom/interface.go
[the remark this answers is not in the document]
mikhailshilkov: > if other resources might need access to this additional metadata

What "other resources"? I personally don't have any idea what other custom resources we are designing for. 

> if we support preview in the future, so it seemed like a more complete abstraction

Again, I'm not sure whether we want to handle previews in extended resources differently from the baseline. I don't think Azure Native or k8s do that.

> should we also be doing more to propegate secret-ness between inputs and outputs more generally in this provider?

How? If input name matches the output name, the engine will do that. If not, how do we know that they are related? But it's worth opening an issue if you have thoughts.

provider/pkg/schema/extension_resource.go
mikhailshilkov: This PR adds 500 lines of net-new code. I consider tests to be a requirement in this case, not a nice-to-have.

Our default mode shall be ~full coverage of all new or meaningfully modified code, and every exception must be motivated.

provider/pkg/schema/extension_resource.go
mikhailshilkov: See my other comment; I don't agree it can be a follow-up

mikhailshilkov:
Thank you for all the changes, I think we are moving forward. I responded to some comments but I want us to take another look at the high level, since I'm still not convinced this is the optimal design.

Each provider CRUD method can roughly be mapped to the following steps:

1. Unmarshal properties from gRPC types to SDK-shaped structures.
2. Map SDK shapes (1) to CF shapes (URN to type name, adjust property casing, unlift properties etc.)
3. Convert wire shapes (2) to CloudControl payloads. This includes patch calculation for Update, for example.
4. Use (3) to interact with CloudControl API, get resource state.
5. Convert resource state (4) to SDK shape (adjust casing, lift to outputs).
6. Calculate resource checkpoint (including input calculation for Read).
7. Marshal to gRPC.

From my point of view, both "normal" and "extension" resources have to follow all these steps. Moreover, steps 1, 3, 4, 6, 7 are exactly the same for them. Step 4 - CC API interaction - is the same, because they can literally be the same AWS resource, as your test demostrates. Also, 3, 4, 6, that are all shared, are the most involved steps with lots of logic and complexity in them.

Only steps 2 and 5 - SDK-to-CF-and-back differ:
1. For normal resources, the type name is defined by resource token and CF payload is driven by metadata.
2. For extension resources, both are taken verbatim from the SDK sub-properties.

Your PR forks steps 1-7 entirely and re-implements them from scratch for the extension resource, hidden behind a new abstraction of a custom resource. In my opinion, this creates unneeded duplication.

Somewhat relatedly, I don't think we have any other candidates for custom resources yet, so we may not have enough data to design the CustomResource abstraction future-proof. It does seem to contribute a meaningful cost with unclear benefit.

> I originally attempted to implement this inline and re-use as much of the code as possible but this didn't go well as many areas need to account for if the properties are directly in the resource or if they're nested inside the "properties" or "outputs" container in the ExtensionResource. This led to lots of special case switches all over the code which made it fragile and complicated. 

I would like to review this approach and see if we can simplify it well enough. It doesn't have to be a bunch of if-else code spread around CRUD methods. In my imagination, we could introduce a new abstraction for the converter steps (2) and (5). One converter implementation would do what the normal resource path, the other one would extract typeName and properties from the payload. Based on your comment, I realize I may not understand the full complexity - so the next logical step is to discuss the code and its problems side-by-side with your current PR. For now, my prior is that we can achieve a much smaller code delta and higher reusability and maintainability with this approach.

> The approach I'd like to take on the duplication is to pull out more of these inline pieces of code into their own functions which both can use instead - but I'm leaving that as a follow-up task, secondary to getting the resource in a working state for launch.

You imply this will happen after the launch. If so, the most likely outcome will be that we will end up with maintaining this new code indefinitely. The price is high so I'm setting a high bar to accept it.

> Also, to note - this isn't just a fix of a single panic - this resource was fundamentally unusable in multiple areas and this is a complete rewrite from scratch of the whole provider implementation for it.

Great, thank you for doing so! Did you have a minimal change that makes your test pass, at some point? It would help me personally to understand the motivation for your work. All things equal, I prefer the sequence "add a test, make it work, then refactor" to complete rewrites.

Let me know if it makes sense and how I can help move it forward.

danielrbradley:
> Did you have a minimal change that makes your test pass, at some point? It would help me personally to understand the motivation for your work. All things equal, I prefer the sequence "add a test, make it work, then refactor" to complete rewrites.

This was exactly the premise I started with - writing the simplest possible integration test and trying to get it to pass. When I raised this for review, that was the first time I'd managed to get that test to pass. This is after several attempts around the design. 

The first was just scattering more if-statements throughout the provider code, but I was still struggling to fix the behaviour without introducing regressions in the schema-based resources. Some of the challenge here is that this resource adds a whole extra layer of nesting - the inputs and outputs are enveloped without the "properties" and "outputs" properties, respectively, so has to have a custom step after unmarshalling and before marshalling to handle this nesting. 

The next challenge is that we don't have metadata which means we need to skip steps 3 and 5. A second approach was to try to construct an on-the-fly schema for this dynamic resource, but that wasn't sufficient so also had to add extra hooks/callbacks within the main process which made the control flow more indirected and much harder to follow. Similarly, the metadata is required for the current diff implementation, which is why a new diff is required for this resource. There was previously no implementation for Update for this resource - it just failed.

Another alternative design I considered, but disregarded as it seemed like a bigger change was to write this single resource using pulumi-go-provider then use the muxer (currently in the bridge) to combine the handwritten elements with the auto-generated elements. This would allow us to use a more well-supported resource abstraction. The downside is that it affects the core entry point of the provider and schem gen programs and could have unintented affects on behaviour. Therefore, the current approach seemed like a smaller change for the time being that could probably be quite easily refactored into the go-provider later if we want.

So the design I've ended up with is the minimal change to the schema-based resource flow to get the simplest test to pass. Happy to build out the test further and add a load more refactoring to de-duplicate bits of copied code, but this is already in a much better state for the extension resource. As long as we're confident this doesn't regress the main resources, this is a clear net-gain already.

mikhailshilkov:
@danielrbradley Here is a diff where tried to illustrate my idea: https://github.com/pulumi/pulumi-aws-native/compare/mikhailshilkov/extension-resource-concept?expand=1

The benefits that I have in mind:
1. Solves scattering more if-statements throughout the provider code.
2. Encapsulates the extra layer of nesting in a single place.
3. Simplifies the provider code itself.
4. Converters are easily testable.
5. Does not duplicate any of CC API interactions.
6. Does not deal with PropertyMap unmarshalling.

My change works fine for Create and Delete. Update and Read are possibly more involved, but can follow the same canvas.

My choice of names may be horrible - I haven't spent any time thinking about them. Error handling is missing too.

I'd also suggest to ship the overall change in multiple PRs:
1. Add a simple one-step test, no tags or write-only annotations, ship Create, Read and Delete implementation.
2. Add Update with a two-step test.
3. Add tags and write-only annotations (honestly, can wait until anybody asks for it).

Does this help at all?

danielrbradley:
I understand what you're trying to do @mikhailshilkov - and wrapping the special case in a converting is a different approach to this. This still doesn't deal with several other concerns including:
- Any resource with write-only properties won't be usable (update and read won't work)
- Default tags can't work
- There wouldn't be any way to implement autonaming if we wanted to (this is still missing from this PR too)

The current state of this PR as I see it is that it comprehensively addresses the failings of the current implementation and makes this resource usable for third part extensions such as DataDog. I've addressed all minor concerns and the only point of contention is the right point to pull out an abstraction to switch between a regular resource and the extension resource. While I could see it might be aguable that a smaller abstraction might be theorically possible, I don't see the value in spending more days refactoring this work further. Adjusting of this design is also very possible later down the line and might be easier once we've address other elements of the design of the overall provider. I would therefore like to merge what's working given the confidence we've got from the integration tests and move on to more pressing concerns. Alternatively, feel free to open up a new PR as you see fit either as an alternative or on top of this PR.

mikhailshilkov:
> This design still doesn't deal with several other concerns including:
> - Any resource with write-only properties won't be usable (update and read won't work)
> - Default tags can't work
> - There wouldn't be any way to implement autonaming if we wanted to (this is still missing from this PR too)
> This will not be able to match the feature set here without several other places to integrate the special casing.

I don't think it's fundamentally limiting, it's just that I haven't implemented those. They can be extracted from inputs and applied to payloads. I suspect it is going to be the same or smaller number of special cases than your current PR.

> The current state of this PR as I see it is that it comprehensively addresses the failings of the current implementation

My very minimal bar is to have ~100% unit test coverage for the net-new code, regardless of the overall design.

> I don't see the value in spending more days refactoring this work further. Adjusting of this design is also very possible later down the line and might be easier once we've address other elements of the design of the overall provider.

The cost of code ownership >> the cost of writing it. I'm curious to get other reviews on this PR, and I may be convinced we want to move this direction, but I don't love the argument of "that's what we have, let's merge it because we don't have time".

danielrbradley:
Writing tests for the sake of tests is worthless - just to say we've got 100% showing up on the metrics. What matters is confidence.

We already know we've got reasonably decent coverage from the integration test. What specific areas do you see need for concern which should stop us merging?

If you've got suggestions for areas where a unit test would cover a critical part on which we must not regress, then that's worth looking at, otherwise this feels like a pointless exercise when we could be closing out other bugs and adding tests around those.

mjeffryes (commented):
Honestly, while I agree with Mikhail that there's probably more that could be extracted out into shared methods; it does actually seem more conservative to me to mostly fork the code paths for Extension Resource, so that we can avoid disrupting the code paths for normal resources.

That's actually my biggest concern with this PR as it stands; it's pretty hard to visually inspect to verify that there are no behavior changes for other resources. (Eg. I like that you've extracted out some new modules, but it creates a lot of visual clutter; I think maybe we could ship that in a separate review first?) 

One challenge we'll have to watch out for is that it can be easy for later changes to add code outside the if statement that doesn't actually apply to both paths, but I think that problem is not really made any better or worse by this change.

Overall, I would be inclined to ship this roughly as is, though ideally in a few smaller chunks and with a few unit tests to help improve our confidence that we're not breaking any existing resources.

examples/untyped-py/step1/__main__.py
```diff
@@ -0,0 +1,36 @@
...
+                             },
+                             write_only=[
+                                 "AllowedPattern",
+                                 "Description",
+                                 "Policies",
+                                 "Tags",
+                                 "Tier"
+                             ],
+                             tags_property="Tags",
+                             tags_style="stringMap",
```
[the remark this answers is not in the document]
mikhailshilkov: We need to make them replaceOnChanges, and the typeName too.

mikhailshilkov:
>  it does actually seem more conservative to me to mostly fork the code paths for Extension Resource, so that we can avoid disrupting the code paths for normal resources.

Should the normal flow be another CustomResource in this model then? Or are we afraid to touch the normal code flow? When we ship ExtensionResource into production, will we be similarly averse about changing its code in any way?

examples/untyped-py/step1/__main__.py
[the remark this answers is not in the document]
danielrbradley: Yes, it's not ideal that users have to provide these, though they are optional so it can be used without them (though write-only properties can result in updates failing if not specified).

To make the most common patterns work without user input, perhaps we should change the default `tagsProperty` to be `Tags` if unspecified (which would be the case for >90% of resources) and default the `tagsStyle` to `keyValueArray` which is also the most common style. In the less common case of wanting to opt-out of default tags the user could then set `tagsStyle` to `none`.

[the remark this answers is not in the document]
danielrbradley: On second thoughts, I think defaulting to blindly adding a tags field is probably surprising behaviour as this could lead to the extension resource payload being invalid - adding a "Tags" field to something that doesn't expect tags. Therefore, I think we need some kind of basic opt-in for applying tags which I'd suggest should be by specifying either the `tagsProperty` or `tagsStyle` - we will the apply the default to the other.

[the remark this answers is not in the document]
danielrbradley: I've added the `replaceOnChanges`, however, due to https://github.com/pulumi/pulumi/issues/15726 this has no effect since we removed all inputs from also being outputs. The gist of the bug is that we actually mark which inputs cause replacements on the output property 🙃 

At some point when the bug it fixed and we upgrade our dependency this will resolve itself.

danielrbradley:
This is now stacked on top of #1442 and squashed to improve clarity of the change as requested.

Since your last review @mjeffryes this now has an additional test & fix for the key value tag style.

provider/pkg/default_tags/merging.go
```diff
@@ -33,6 +33,12 @@ func MergeDefaultTags(tags resource.PropertyValue, defaultTags map[string]string
 
 	if tagsStyle.IsKeyValueArray() {
 		var asArray []resource.PropertyValue
+		keyProp := resource.PropertyKey("key")
+		valueProp := resource.PropertyKey("value")
+		if tagsStyle == TagsStyleKeyValueArrayUpperCase {
+			keyProp = resource.PropertyKey("Key")
+			valueProp = resource.PropertyKey("Value")
+		}
```
mjeffryes, line 41: hrmmm... this delta is a bummer since it seems to just be a matter of skipping name normalization for the extension resource. I suppose normalizing names would be an even more messy change though.  Would like to see a unit test for this though.

provider/pkg/resources/extension_resource.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func NewExtensionResource(client client.CloudControlClient) *extensionResource {
+	return &extensionResource{
+		client: client,
+	}
+}
+
+func (r *extensionResource) Check(ctx context.Context, urn resource.URN, inputs, state resource.PropertyMap, defaultTags map[string]string) (resource.PropertyMap, []ValidationFailure, error) {
+	var extensionResource ExtensionResource
```
mjeffryes, line 37: nit. I find the variable name and two type names all being [Ee]xtensionResource to be kind of confusing to read.

provider/pkg/provider/provider.go
```diff
@@ -1035,46 +1037,73 @@ func (p *cfnProvider) Update(ctx context.Context, req *pulumirpc.UpdateRequest)
...
+		Label:        fmt.Sprintf("%s.oldState", label),
+		KeepUnknowns: true,
+		RejectAssets: true,
+		KeepSecrets:  true,
+	})
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "diff failed because malformed resource inputs")
 	}
+	oldInputs := parseCheckpointObject(oldState)
```
mjeffryes, line 1049: Looks like this is new behavior on existing resources not just the Extension Resource; curious about the motivation for adding it.

mjeffryes (approved):
couple nits, but I think this is reasonable to ship.

provider/pkg/provider/provider.go
danielrbradley: Ah yes, this is currently unused on the schema resource condition. This is because the `p.diffState()` method takes the grpc type, the calls parseCheckpointObject itself.

I avoided initially touching the `diffState` method, but think this warants a clean-up to avoid doing the work twice. This means that we can completely remove `diffState` as it's literally just doing the unpacking of the `oldInputs` then calling diff.

danielrbradley:
> couple nits, but I think this is reasonable to ship.

Thanks @mjeffryes! I've pushed 3 new commits each of which addresses your comments:


1. [Rename ExtensionResource types & vars](https://github.com/pulumi/pulumi-aws-native/pull/1406/commits/9298a60cbde6384e817b7d7cfc17210e11833251) 
2. [Add default tags uppercase test](https://github.com/pulumi/pulumi-aws-native/pull/1406/commits/76e06fd08eab28f8ba7e787c3ffb78f3ae950dde)
3. [Remove diffState helper](https://github.com/pulumi/pulumi-aws-native/pull/1406/commits/2cc865247f8738d16a2f8c1862b6cccc98330ef3)

PR #1409, opened 2024-03-13
Add a thin wrapping layer around the CC SDK calls

mikhailshilkov:
Inspired by the change by @thomas11 in Azure Native, I'm removing the direct dependency from the provider code to the AWS CloudControl SDK and adding a thin wrapper layer in-between. I also moved the await logic into a separate file.

The main purpose of this change is to make the provider code unit-testable with the CC client mocked out. It's also a first step towards adding more modularity around the await and error handling logic.

No functional changes in this PR.

### DIFF
--- a/provider/pkg/provider/awsclient/awaiter.go
+++ b/provider/pkg/provider/awsclient/awaiter.go
@@ -1,0 +1,96 @@
+// Copyright 2016-2024, Pulumi Corporation.
+
+package awsclient
+
+import (
+	"context"
+	"time"
+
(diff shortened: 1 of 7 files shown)


danielrbradley:
Snap! I've got the exact same change in my PR to fix the ExtensionResource: https://github.com/pulumi/pulumi-aws-native/pull/1406/commits/8aaba951e830bad15d8ef226c063df9ea0f374a9 😭

danielrbradley:
One notable difference on review is that I've made the await logic internal to the client so this compexity is also removed from the provider.

danielrbradley (approved):

PR #1416, opened 2024-03-14
Add a Client abstraction

mikhailshilkov:
Adds another client layer around Cloud Control operations to encapsulate serialization, error handling, and awaiting long-running operations.

No functional changes.

### DIFF
--- a/provider/pkg/provider/client/client.go
+++ b/provider/pkg/provider/client/client.go
@@ -14,10 +14,23 @@ import (
 	"github.com/pulumi/pulumi-aws-native/provider/pkg/schema"
 )
 
-type Client interface {
+// CloudControlApiClient providers CRUD operations around Cloud Control API, with the mechanics of API calls abstracted away.
+// For instance, it serializes and deserializes wire data and follows the protocol of long-running operations.
+type CloudControlClient interface {
+	// Create creates a resource of the specified type with the desired state.
+	// It awaits the operation until completion and returns a map of output property values.
 	Create(ctx context.Context, typeName string, desiredState map[string]any) (identifier *string, resourceState map[string]any, err error)
+
+	// Read returns the current state of the specified resource. It deserializes
+	// the response from the service into a map of untyped values.
 	Read(ctx context.Context, typeName, identifier string) (map[string]interface{}, error)
+
+	// Update updates a resource of the specified type with the specified changeset.
+	// It awaits the operation until completion and returns a map of output property values.
 	Update(ctx context.Context, typeName, identifier string, patches []jsonpatch.JsonPatchOperation) (map[string]interface{}, error)
+
+	// Delete deletes a resource of the specified type with the given identifier.
+	// It awaits the operation until completion.
 	Delete(ctx context.Context, typeName, identifier string) error
 }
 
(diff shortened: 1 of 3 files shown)


provider/pkg/provider/client/client.go
```diff
@@ -0,0 +1,130 @@
...
+			return nil, nil, waitErr
+		}
+
+		if readErr != nil {
+			return nil, nil, fmt.Errorf("resource partially created but read failed. read error: %v, create error: %w", readErr, waitErr)
+		}
+
+		// Resource was created but failed to fully initialize.
+		// If it has some state, return a partial error.
+		return &id, outputs, waitErr
```
danielrbradley, line 79: Should we be checking for the id on this branch too so that `id` and `outputs` are both always nill or not nill?

provider/pkg/provider/provider.go
```diff
@@ -837,46 +836,16 @@ func (p *cfnProvider) Create(ctx context.Context, req *pulumirpc.CreateRequest)
...
-	glog.V(9).Infof("%s.CreateResource %q token %q state %q", label, cfType, clientToken, desiredState)
-	res, err := p.cctl.CreateResource(ctx, &cloudcontrol.CreateResourceInput{
-		ClientToken:  aws.String(clientToken),
-		TypeName:     aws.String(cfType),
-		DesiredState: aws.String(desiredState),
-		RoleArn:      p.roleArn,
-	})
-	if err != nil {
+	glog.V(9).Infof("%s.CreateResource %q", label, cfType)
+	id, resourceState, err := p.client.Create(ctx, cfType, payload)
```
danielrbradley, line 841: I think this error needs to be bound to a separate variable so we can inspect it later to decide if we're returning a partialCreate error – so the error doesn't get overwritten by a serialization error.

danielrbradley (commented):
The shape of the client looks good. Just a question on how we're handling the partial create case.

provider/pkg/provider/provider.go
mikhailshilkov: Yes, great catch

provider/pkg/provider/client/client.go
mikhailshilkov: I think `outputs` are never nil here? They are assigned if `readErr` is nil, and `readErr` is known to be nil by this line.

So, both `id` and `outputs` are defined here, and we return `nil, nil` everywhere else. What am I missing?

provider/pkg/provider/client/client.go
```diff
@@ -0,0 +1,130 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudcontrol"
+	"github.com/aws/aws-sdk-go-v2/service/cloudcontrol/types"
+	"github.com/mattbaird/jsonpatch"
+	"github.com/pkg/errors"
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/schema"
+)
+
+type Client interface {
```
thomas11, line 17: The name "Client" doesn't say much because there are many clients. I'd be more specific, but it's subjective, of course.

thomas11 (approved):
nice!

provider/pkg/provider/client/client.go
mikhailshilkov: @danielrbradley Are you okay to check this to CloudControlClient in your branch? I'd prefer that name to Client too.

danielrbradley: Sure - I agree that name would be better

danielrbradley (approved):

PR #1429, opened 2024-03-15
Remove fixed policy name in the Py default tags test

mikhailshilkov:
It seems to be causing a concurrency failure https://github.com/pulumi/pulumi-aws-native/actions/runs/8294834587/job/22700959209#step:25:266

Fixes https://github.com/pulumi/pulumi-aws-native/issues/1428


danielrbradley (approved):
Beat me to it while looking into the P1

mikhailshilkov:
Opened https://github.com/pulumi/pulumi-aws-native/issues/1430 as a follow-up

PR #1431, opened 2024-03-15
Client refactor

danielrbradley:
Sort out dependencies so we can introduce #1406 without introducing cycles.

1. Move client to top-level. If it's within provider then we can't import it into another module that's going to be imported by provider.

2. Remove resource state conversion in client. This is also done in the provider and should not be a concern of the client.
This also allows us to remove a dependency on pkg/schema.


provider/pkg/client/client.go
```diff
@@ -68,17 +67,14 @@ func (c *clientImpl) Create(ctx context.Context, typeName string, desiredState m
...
 	if pi != nil && pi.Identifier != nil {
 		// Retrieve the resource state from AWS.
 		// Note that we do so even if creation hasn't succeeded but the identifier is assigned.
 		id = *pi.Identifier
-		resourceState, err := c.api.GetResource(ctx, typeName, id)
+		resourceState, err = c.api.GetResource(ctx, typeName, id)
 		if err != nil {
 			readErr = fmt.Errorf("reading resource state: %w", err)
-		} else {
-			outputs = schema.CfnToSdk(resourceState)
```
mikhailshilkov, line 81: Oh, that was my mistake! Thank you for noticing!

mikhailshilkov (approved):
LGTM

mikhailshilkov (changes requested):
Sorry, I think you are missing a corresponding CfnToSdk call in the provider?

mikhailshilkov (approved):
No, you are all good, it was duplicated. Sorry for the noise.

Issue #1304, opened 2024-01-31
Panic converting a synthesized basic CDK program


jkodroff:
### What happened?

I get a panic converting a basic synthesized CDK program:

```bash
$ cf2pulumi nodejs ../cdk/template.yaml
panic: fatal: An assertion has failed

goroutine 1 [running]:
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /Users/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.91.1/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Assert(...)
        /Users/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.91.1/go/common/util/contract/assert.go:28
github.com/pulumi/pulumi-aws-native/provider/pkg/cf2pulumi.(*renderContext).renderResource(0x140030f0eb0, 0x14002a67bf0?)
        /Users/runner/work/pulumi-aws-native/pulumi-aws-native/provider/pkg/cf2pulumi/renderer.go:871 +0x11d4
github.com/pulumi/pulumi-aws-native/provider/pkg/cf2pulumi.RenderTemplate(0x16b9a2e09?, 0x140032ba2e8)
        /Users/runner/work/pulumi-aws-native/pulumi-aws-native/provider/pkg/cf2pulumi/renderer.go:1296 +0x1138
github.com/pulumi/pulumi-aws-native/provider/pkg/cf2pulumi.RenderFile({0x16b9a2e09, 0x14}, 0x70c83?)
        /Users/runner/work/pulumi-aws-native/pulumi-aws-native/provider/pkg/cf2pulumi/renderer.go:1317 +0xc4
main.main()
        /Users/runner/work/pulumi-aws-native/pulumi-aws-native/provider/cmd/cf2pulumi/main.go:46 +0x1e4
```

### Example

Here's my CDK source:

```typescript
import * as cdk from 'aws-cdk-lib';
import { Vpc } from 'aws-cdk-lib/aws-ec2';
import { Construct } from 'constructs';

export class CdkStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const vpc = new Vpc(this, "convert-cfn", {
      natGateways: 1,
      restrictDefaultSecurityGroup: false
    });
  }
}
```

[...]

PR #1438, opened 2024-03-18
Improve cf2pulumi error messages

danielrbradley:
Remove all uses of contract.Assert - return errors where possible or otherwise use Assertf with additional contextual information.

Related to https://github.com/pulumi/pulumi-aws-native/issues/1304


mjeffryes (approved):
Thanks for chasing down all the asserts; much better to address 10 possible panics than just one!

PR #1442, opened 2024-03-19
Refactor modules

danielrbradley:
- Move lots out from schema into smaller modules with less dependencies.
- Fix a few compiler warnings along the way.
- Rename the gen `context` struct so it doesn't block us using the `context` std lib.
- Fix flakey lambda update integration test.

This was extracted from https://github.com/pulumi/pulumi-aws-native/pull/1406

### DIFF
--- a/examples/lambda-update/step1/index.ts
+++ b/examples/lambda-update/step1/index.ts
@@ -31,7 +31,6 @@ export const graphqlPublicZip = new aws.s3.BucketObject(
   "lambda-source",
   {
     bucket: lambdaStore.id,
-    key: "samplefn",
     source: new pulumi.asset.AssetArchive({
       "index.js": new pulumi.asset.StringAsset(
         `module.exports = () => { console.log("placeholder"); return Promise.resolve(); }`
@@ -46,7 +45,6 @@ export const layerCode = new aws.s3.BucketObject(
   "layer-source",
   {
     bucket: lambdaStore.id,
-    key: "samplefn",
     source: new pulumi.asset.AssetArchive({
       "index.js": new pulumi.asset.StringAsset(
         `module.exports = () => { console.log("placeholder"); return Promise.resolve(); }`
--- a/provider/pkg/resources/validate.go
+++ b/provider/pkg/resources/validate.go
@@ -1,15 +1,17 @@
 // Copyright 2016-2021, Pulumi Corporation.
 
-package schema
+package resources
 
 import (
 	"fmt"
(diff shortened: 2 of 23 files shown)


examples/lambda-update/step1/index.ts
```diff
@@ -31,7 +31,6 @@ export const graphqlPublicZip = new aws.s3.BucketObject(
   "lambda-source",
   {
     bucket: lambdaStore.id,
-    key: "samplefn",
```
mjeffryes, line 34: Why the change to this example?

provider/pkg/resources/validate.go
```diff
@@ -1,15 +1,17 @@
 // Copyright 2016-2021, Pulumi Corporation.
 
-package schema
+package resources
 
 import (
 	"fmt"
+	"math"
+	"strings"
```
mjeffryes, line 8: I don't see where these we're added in the code. Were they just missing before?

provider/pkg/schema/gen.go
```diff
@@ -1442,3 +1446,5 @@ func GatherSemantics(schemaDir string) (SemanticsSpecDocument, error) {
 	}
 	return semanticsDocument, nil
 }
+
+var ExtensionResourceToken string = "aws-native:index:ExtensionResource"
```
mjeffryes, line 1450: I don't see where this moved from?

mjeffryes (approved):
Few nits, but looks good overall. Thanks for splitting this out.

examples/lambda-update/step1/index.ts
danielrbradley: Both BucketObjects were created with the same key in the same bucket, so one overwrote the other. Removing the key means we'll just use auto-naming based on the Pulumi identifier. 

This was surfacing as a refresh diff sometimes during the test - where the version id of the object had changed. Essentially this is multiple data races: which one gets written first or second, then once written if one gets read before or after the other has also been written - and therefore if it actually gets a different version to what it puts 🤦

mjeffryes: Whoops that is bad. Thanks for clarifying!

provider/pkg/schema/gen.go
danielrbradley: Ah, it's in the metadata.go .. and it's still there at this point, then it should be removed in the next PR. I also couldn't find where it had gone so fixed the compile error with this line.

PR #1443, opened 2024-03-19
Improve autonaming

### LINKED ISSUE #1430: resiliencehub.ResiliencyPolicy does not apply auto-naming
### What happened?

The code snippet below fails with `policy_name` is required. I expect `policy_name` to be optional and to be calculated from the resource name by default (auto-naming).

### Example

```py
policy = resiliencehub.ResiliencyPolicy(
  "policy",
  tier=resiliencehub.ResiliencyPolicyTier.NON_CRITICAL,
  policy={
    'Software': resiliencehub.ResiliencyPolicyFailurePolicyArgs(
      rto_in_secs=10,
      rpo_in_secs=10,
    ),
    'Hardware': resiliencehub.ResiliencyPolicyFailurePolicyArgs(
      rto_in_secs=10,
      rpo_in_secs=10,
    ),
    'AZ': resiliencehub.ResiliencyPolicyFailurePolicyArgs(
      rto_in_secs=10,
      rpo_in_secs=10,
    ),
    'Region': resiliencehub.ResiliencyPolicyFailurePolicyArgs(
      rto_in_secs=10,
      rpo_in_secs=10,
    ),
  },
  tags={"localTag": "localTagValue"},
)
```

### Output of `pulumi about`

aws-native version 0.99.0

### Additional context

_No response_

### Contributing

danielrbradley:
- Introduce report to highlight and help investigate resources we can't auto-name.
- Identify partial overlaps of resource and name fields e.g. ScalingPolicy has a field "PolicyName". This fixes 28 resources.
- Fix #1430 by removing hard coded resilience policy name.

There are still currently 387 resources remaining without auto-naming. Many of these genuinely can't be auto-named.

### DIFF
--- a/provider/pkg/autonaming/spec.go
+++ b/provider/pkg/autonaming/spec.go
@@ -7,39 +7,58 @@ import (
 
 	jsschema "github.com/pulumi/jsschema"
 	"github.com/pulumi/pulumi-aws-native/provider/pkg/metadata"
-	"github.com/pulumi/pulumi-aws-native/provider/pkg/naming"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 )
 
 func CreateAutoNamingSpec(inputProperties map[string]pschema.PropertySpec, resourceTypeName string, jsonSchemaProperties map[string]*jsschema.Schema, semanticsSpec metadata.SemanticsSpec) *metadata.AutoNamingSpec {
-	tryMakeSpecForField := func(cfPropName string) (spec *metadata.AutoNamingSpec) {
-		if prop, has := inputProperties[naming.ToSdkName(cfPropName)]; !has || prop.Type != "string" {
+	type names struct {
+		cfName  string
+		sdkName string
+	}
+	// Lowercase all keys so we can do case-insensitive lookups.
+	fieldsLowered := make(map[string]names, len(jsonSchemaProperties))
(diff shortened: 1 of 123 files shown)


mjeffryes (commented):
Looks reasonable; can you add/update a test for `createAutoNamingSpec`?

provider/pkg/autonaming/spec.go
```diff
@@ -0,0 +1,83 @@
...
+			autoNameSpec = potentialSpecs[0]
+		}
+	}
+
+	if autoNameSpec != nil {
+		namingTriviaSpec, ok := semanticsSpec.NamingTriviaSpec[autoNameSpec.SdkName]
+		if ok {
+			autoNameSpec.TriviaSpec = &namingTriviaSpec
+		}
+	}
```
mjeffryes, line 80: tiny style nit, if you move this into `tryMakeSpecForField`, you could use early returns above to avoid nesting the rest of the code in if blocks.

mjeffryes (approved):
Thanks! a couple more test cases that would be good to add but otherwise LGTM.

provider/pkg/autonaming/spec.go
danielrbradley: Ha that's where it started! Though, having changed the outer design, I agree that being able to use the early return would now be nice.

PR #1453, opened 2024-03-22
ExtensionResource autonaming

danielrbradley:
- Move autonaming provider code into autonaming module along with test.
- Add wrapper function to reduce code at call site.
- Add AutoNaming schema for ExtensionResource.
- Update integration test to utilize autonaming.


provider/pkg/provider/provider.go
```diff
@@ -733,14 +734,7 @@ func (p *cfnProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*
...
 
-		if autoNamingSpec := spec.AutoNamingSpec; autoNamingSpec != nil {
-			// Auto-name fields if not already specified
-			val, err := getDefaultName(req.RandomSeed, urn, autoNamingSpec, olds, newInputs)
-			if err != nil {
-				return nil, err
-			}
-			newInputs[resource.PropertyKey(autoNamingSpec.SdkName)] = val
-		}
+		autonaming.ApplyAutoNaming(spec.AutoNamingSpec, urn, req.RandomSeed, olds, newInputs)
```
mjeffryes, line 737: I think there's a behavior change here: the old code returned an error if there was an error from `getDefaultName` but it looks like you're dropping it here.

provider/pkg/resources/extension_resource.go
```diff
@@ -49,6 +58,14 @@ func (r *extensionResource) Check(ctx context.Context, urn resource.URN, inputs,
 		return nil, failures, nil
 	}
 
+	if typedInputs.AutoNaming != nil {
+		autonaming.ApplyAutoNaming(&metadata.AutoNamingSpec{
```
mjeffryes, line 62: should we return the error if this fails?

mjeffryes (commented):
This looks mostly fine, but I'm a little unsure what the motivation is? From the integration test, it's a pretty minor difference, not necessarily easier, just different.

provider/pkg/provider/provider.go
danielrbradley: Great spot! Thank you 👍

danielrbradley:
> This looks mostly fine, but I'm a little unsure what the motivation is? From the integration test, it's a pretty minor difference, not necessarily easier, just different.

It's not a massive difference - was just a wart I noticed while writing the integration test. It's a bit of a pain having to import the whole random provider just to fix naming clashes, and removes the need for an extra resource, when this is why we have auto-naming everywhere else in Pulumi.

mikhailshilkov (commented):
I don't see any tests that verify anything about the auto-naming behavior (other than it doesn't entirely fail).
Also, similar to Matt's comment, I'm not sure why we are doing this. The more metadata properties we add to this resource, the more unique and unexpected it may look for users. Do we have any signal this is useful and requested?

mjeffryes:
> I don't see any tests that verify anything about the auto-naming behavior (other than it doesn't entirely fail). Also, similar to Matt's comment, I'm not sure why we are doing this. The more metadata properties we add to this resource, the more unique and unexpected it may look for users. Do we have any signal this is useful and requested?

I had the same thought, but it does look like the core of autonaming is tested here: https://github.com/pulumi/pulumi-aws-native/blob/88bb457553795978a0a9f4851b36e092f711ee3e/provider/pkg/autonaming/application_test.go#L14

danielrbradley:
> I don't see any tests that verify anything about the auto-naming behavior (other than it doesn't entirely fail). 

That's already tested, this PR just moves the code to make it accessible. The integration test then just asserts that the same functionality is called correctly from the ExtensionResource.

>Also, similar to Matt's comment, I'm not sure why we are doing this. The more metadata properties we add to this resource, the more unique and unexpected it may look for users. Do we have any signal this is useful and requested?

Similar to https://github.com/pulumi/pulumi-aws-native/issues/1430, the original integration test for extension resources had the issue that names were not unique. The quick hack was to use the random provider, but this is proper solution. Ideally, any Cloud Control resource that's able to be used via the SDK should also work in the same way through the ExtensionResource - so it's a nice aim to have feature and behaviour parity between normal resources and the ExtensionResource.

danielrbradley:
Can we unblock this? What's the downside to merging this?

Another benefit of using autonaming over the random provider for this is that it then supports the create before delete use case which would be much more challenging otherwise.

corymhall (approved):

PR #1495, opened 2024-04-19
Add provider to path for running integration tests

flostadler:
Previously you had to manually add the provider to the PATH for the tests to work. This automates that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,7 @@ install_java_sdk::
 install_nodejs_sdk::
 	yarn link --cwd $(WORKING_DIR)/sdk/nodejs/bin
 
+test:: PATH := $(WORKING_DIR)/bin:$(PATH)
 test::
 	cd examples && go test -v -tags=all -timeout 2h
 


danielrbradley (approved):
This looks a good step forward. I would suggest that a more robust solution here is to specify the explit, fully qualified path to where we expect the binary to be in the test options. This then fails the test if the binary is not where we expect rather than falling back to downloading another version or locating another random binary that also happens to be on the path.

Example in Azure Native:
https://github.com/pulumi/pulumi-azure-native/blob/d422a4264e04244bd42a188b798e4c72fec7fa73/examples/examples_test.go#L26-L46

flostadler:
> This looks a good step forward. I would suggest that a more robust solution here is to specify the explit, fully qualified path to where we expect the binary to be in the test options. This then fails the test if the binary is not where we expect rather than falling back to downloading another version or locating another random binary that also happens to be on the path.
> 
> Example in Azure Native: https://github.com/pulumi/pulumi-azure-native/blob/d422a4264e04244bd42a188b798e4c72fec7fa73/examples/examples_test.go#L26-L46

That's a good point, thanks for bringing it up! I think there's some aligning we need to do across the providers because some of them (at least the AWS ones) are missing a lot of those utility functions. I'm gonna have a look at that

PR #1510, opened 2024-05-01
Disambiguate QBusiness::Index Status enum

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws-native/issues/1506

IndexStatus enum name was creating a generate-phase panic as this token was already taken.

### DIFF
--- a/provider/pkg/schema/gen.go
+++ b/provider/pkg/schema/gen.go
@@ -1225,6 +1225,12 @@ func (ctx *cfSchemaContext) genEnumType(enumName string, propSchema *jsschema.Sc
 	case "networkfirewall:RuleGroupType":
 		typName = "RuleGroupTypeEnum" // Go SDK name conflict vs. RuleGroup resource
 	}
+
+	switch ctx.cfTypeName + "   " + enumName {
+	case "AWS::QBusiness::Index   Status":
+		typName = "QBusinessIndexStatus"
+	}
+
 	tok := fmt.Sprintf("%s:%s:%s", packageName, ctx.mod, typName)
 
 	enumSpec := &pschema.ComplexTypeSpec{


provider/pkg/schema/gen.go
```diff
@@ -1225,6 +1225,12 @@ func (ctx *cfSchemaContext) genEnumType(enumName string, propSchema *jsschema.Sc
 	case "networkfirewall:RuleGroupType":
 		typName = "RuleGroupTypeEnum" // Go SDK name conflict vs. RuleGroup resource
 	}
+
+	switch ctx.cfTypeName + "   " + enumName {
```
flostadler, line 1229: Why is this specifically 3 whitespaces?

provider/pkg/schema/gen.go
```diff
@@ -1225,6 +1225,12 @@ func (ctx *cfSchemaContext) genEnumType(enumName string, propSchema *jsschema.Sc
...
+	case "AWS::QBusiness::Index   Status":
+		typName = "QBusinessIndexStatus"
```
flostadler, line 1231: This could happen at any point again, right? Is there a generic approach we could implement to prevent this from happening again in the future? I guess namespacing (like adding the service name in front) the enum types now would be a breaking change, right?

mjeffryes (approved):

flostadler (approved):

t0yv0:
The general purpose fix is moving these to a CSV file from the code. One day.

provider/pkg/schema/gen.go
t0yv0: No reason, just need some sort of separator since I can't pattern match a  tuple (a, b) against a pattern (a, b) in Go.

t0yv0: This an odd thing to test since it's not quite a feature, it's a data correction. The schemagen decides how to map cfTypeName to Pulumi type token with some degrees of freedom, and we're adding some configuration here to direct it better. I think it will be obvious if this works in subsequent PR that should avoid the panic. If this happens often we should make this a true feature of the generator where it can source config data that informs this name mapping, and test that.

Issue #1336, opened 2024-02-10
APIs are missing some documentation


joeduffy:
I'm unsure if this is intentional, but a number of the resources have missing property documentation. For example, this type is missing it entirely (https://www.pulumi.com/registry/packages/awsx/api-docs/ecs/fargateservice/#taskdefinitionlogconfiguration) but if you scroll around the page, there are partly documented types too.

mikhailshilkov:
@mjeffryes It looks like the aws-native provider descriptions are fixed in this particular place, see https://www.pulumi.com/registry/packages/aws-native/api-docs/ecs/taskdefinition/#taskdefinitionlogconfiguration

Is it a matter of releasing a new version of awsx now to fix the issue reported by Joe?

Also, let's either find a new example of missing descriptions and update this issue, or close it if the problem is truly fixed now.

PR #1542, opened 2024-05-16
chore: augment documentation with CloudFormation docs

### LINKED ISSUE #63: Descriptions are missing in API docs
Our current API docs are missing textual descriptions for resource and property descriptions. We put API doc links instead which won't lead to a good API doc user experience. We need to retrieve the docs instead.

corymhall:
The CDK team publishes a JSON file with the contents of the
CloudFormation documentation. It can be found here
https://github.com/cdklabs/awscdk-service-spec/tree/main/sources/CloudFormationDocumentation

For whatever reason, the CloudFormation schema does not include a lot of
documentation. This PR uses the CloudFormation docs to augment the
missing documentation. This PR is the first pass at this.

It also includes a report that reports on how many properties have been
augmented and how many still have missing docs. There are still some
that are missing, but I think the number it does populate is pretty good
for a first pass.

```
Number of docs augmented: 14050
Number of docs missing: 818
```

fixes #63, #1336


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -34,7 +34,7 @@ import (
 
 func main() {
 	flag.Usage = func() {
-		const usageFormat = "Usage: %s <operation> <schema-folder> <version> (<schema urls>)"
+		const usageFormat = "Usage: %s <operation> <schema-folder> <version> (<schema urls>) <docs-url>"
```
t0yv0: A lot of positional arguments, could we use keyword based `-docs-url` flag or some such?

provider/pkg/schema/docs_schema.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return &docs, nil
+}
+
+func (d *Docs) GetPropertyDesc(typeName, propName string) (string, bool) {
+	if typ, exists := d.Types[typeName]; exists {
+		if propName == "" {
+			return typ.Description, true
+		}
+		if prop, ok := typ.Properties[propName]; ok {
```
t0yv0, line 41: Curious, no difference between Properties and Attributes for the purposes here? It always prefers Properties?

provider/pkg/schema/docs_schema.go
```diff
@@ -0,0 +1,49 @@
...
+	dec := json.NewDecoder(in)
+	if err := dec.Decode(&docs); err != nil {
+		return nil, fmt.Errorf("error decoding: %v", err)
+	}
+	return &docs, nil
+}
+
+func (d *Docs) GetPropertyDesc(typeName, propName string) (string, bool) {
+	if typ, exists := d.Types[typeName]; exists {
+		if propName == "" {
```
t0yv0, line 38: This is a little surprising compared to a separate `GetTypeDec(typeName) (string, bool)`

provider/pkg/schema/gen_test.go
```diff
@@ -0,0 +1,334 @@
...
...iner. You can specify up to ten environment files. The file must have a `.env` file extension. Each line in an environment file should contain an environment variable in `VARIABLE=VALUE` format. Lines beginning with `#` are treated as comments and are ignored. For more information about the environment variable file syntax, see [Declare default environment variables in file](https://docs.aws.amazon.com/https://docs.docker.com/compose/env-file/) .\n\nIf there are environment variables specified using the `environment` parameter in a container definition, they take precedence over the variables contained within an environment file. If multiple environment files are specified that contain the same variable, they're processed from the top down. We recommend that you use unique variable names. For more information, see [Specifying environment variables](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html) in the *Amazon Elastic Container Service Developer Guide* .\n\nThis parameter is only supported for tasks hosted on Fargate using the following platform versions:\n\n- Linux platform version `1.4.0` or later.\n- Windows platform version `1.0.0` or later.",
```
t0yv0, line 278: Consider to offload to the testdata folder, easy to achieve with autogold package .

t0yv0:
This looks like a big win, the code is fine, do we want to review the docu content in some form? Spot-check? Does it have some references to CloudFormation that need to be edited or massaged out?

sdk/dotnet/AmplifyUiBuilder/Inputs/FormFileUploaderFieldConfigArgs.cs
```diff
@@ -14,24 +14,45 @@ public sealed class FormFileUploaderFieldConfigArgs : global::Pulumi.ResourceArg
...
         }
 
+        /// <summary>
+        /// The access level to assign to the uploaded files in the Amazon S3 bucket where they are stored. The valid values for this property are `private` , `protected` , or `public` . For detailed information about the permissions associated with each access level, see [File access levels](https://docs.aws.amazon.com/https://docs.amplify.aws/lib/storage/configureaccess/q/platform/js/) in the *Amplify documentation* .
+        /// </summary>
         [Input("accessLevel", required: true)]
         public Input<Pulumi.AwsNative.AmplifyUiBuilder.FormStorageAccessLevel> AccessLevel { get; set; } = null!;
 
+        /// <summary>
+        /// Allows the file upload operation to be paused and resumed. The default value is `false` .
```
t0yv0, line 34: Nice

t0yv0:
Should we vendor or retain an S3 copy of the sources just in case upstream disappears.

provider/pkg/schema/docs_schema.go
corymhall: I'll create a separate issue to track this. There was some additional complexity for figuring out which were attributes and which were properties, and the attributes in the CFN docs were structured differently.

There is a chance that some properties and attribute overlap at the same path in the schema and then we may have the wrong description for the time being.

provider/pkg/schema/docs_schema.go
corymhall: What do you mean? Split this into two functions?

corymhall:
> This looks like a big win, the code is fine, do we want to review the docu content in some form? Spot-check? Does it have some references to CloudFormation that need to be edited or massaged out?

Later in the process when we covert to the pschema we do some description editing.
https://github.com/pulumi/pulumi-aws-native/blob/56e4be269e4bc3bef537d64e08a4612435a1fdfa/provider/pkg/schema/gen.go#L902

> Should we vendor or retain an S3 copy of the sources just in case upstream disappears.

Maybe? I thought about checking it in, but it would add another 10MB (at this point maybe we don't care about that?). I think there is a low chance of upstream disappearing similar to how we don't expect `terraform-provider-aws` to disappear.

provider/pkg/schema/docs_schema.go
t0yv0: Yeah, I would if it were a public API. This is just a stylistic nit, this would matter if we were building a library for pubilc use but internally I guess it just doesn't matter, feel free to disregard.

t0yv0:
We can just fork the repo into pulumi org that's probably sufficient.

provider/cmd/pulumi-gen-aws-native/main.go
corymhall: ok I refactored this to use flags instead of arguments

provider/pkg/schema/gen.go
```diff
@@ -892,11 +917,30 @@ func addUntypedPropDocs(propertySpec *pschema.PropertySpec, cfTypeName string) {
...
+	}
+}
+
 func (ctx *cfSchemaContext) propertySpec(propName, resourceTypeName string, spec *jsschema.Schema) (*pschema.PropertySpec, error) {
 	typeSpec, err := ctx.propertyTypeSpec(naming.LowerAcronyms(propName), spec)
 	if err != nil {
 		return nil, err
 	}
+
+	ctx.reportMissingDocs(spec.Description, propName)
```
flostadler, line 942: nit: It might be a bit surprising that it only reports the missing docs if description is empty.

But that's very nit and just my own preference so feel free to ignore

provider/pkg/schema/gen.go
```diff
@@ -916,6 +960,117 @@ func (ctx *cfSchemaContext) propertySpec(propName, resourceTypeName string, spec
 	return &propertySpec, nil
 }
 
+// updateDesc updates the schema with a description from the CloudFormation docs if one
+// is not already provided
+func (ctx *cfSchemaContext) updateDesc(refName, propName string, spec *jsschema.Schema) {
```
flostadler, line 965: what if the existing docs are worse than the new docs? I wonder if we need some way to select between the two eventually in the future

provider/pkg/schema/gen.go
```diff
@@ -805,12 +820,21 @@ func (ctx *cfSchemaContext) gatherResourceType() error {
...
 		originalSdkName := sdkName
 		if sdkName == "id" {
 			sdkName = "awsId"
 		}
+
+		// augment the documentation with the CloudFormation documentation
+		// we do this before creating the propertySpec because in the process of
+		// creating the propertySpec we change some of the names of properties which
+		// makes it impossible to lookup in the docs
+		ctx.augmentDocumentation(ctx.cfTypeName, prop, spec)
```
flostadler, line 836: Are we overloading the `markCreateOnlyProperties` func with too much functionality here? I wouldn't expect that the docs are enhanced within this func

flostadler (approved):
I spot checked ~30 resources across the different SDKs and couldn't spot any fallout. Looks good to me!

sdk/dotnet/IoT/ThingGroup.cs
```diff
@@ -15,15 +15,31 @@ namespace Pulumi.AwsNative.IoT
...
+        /// 
+        /// A Dynamic Thing Group does not have `parentGroupName` defined.
+        /// </summary>
         [Output("parentGroupName")]
         public Output<string?> ParentGroupName { get; private set; } = null!;
 
+        /// <summary>
+        /// The dynamic thing group search query string.
+        /// 
+        /// The `queryString` attribute *is* required for `CreateDynamicThingGroup` . The `queryString` attribute *is not* required for `CreateThingGroup` .
```
t0yv0, line 41: NIt: `queryString` does not inflect into a .NET reference.

t0yv0 (approved):
CC @interurban

provider/pkg/schema/gen.go
corymhall: I think that is the only place that docs are, unless I am missing something.

provider/pkg/schema/gen.go
corymhall: Yeah that's a good idea, i'll create an issue to track it.

provider/pkg/schema/gen.go
corymhall: I think that's just how the diff folds are displayed. It's part of the `gatherResourceType` function.

PR #1572, opened 2024-06-05
Fix truncated descriptions in docs

### LINKED ISSUE #1550: Ability to overwrite existing property docs with CloudFormation docs
## Hello!

## Issue details

We currently augment missing docs with CloudFormation docs, but we might want to create the ability to selectively pick the CloudFormation docs if those are better.

flostadler:
The documentation of some resources/properties is incomplete or incorrect in the AWS CloudFormation schema.
The CloudFormation docs are more complete and accurate in these cases. This change adds the ability to choose
for which resources the CloudFormation docs should be preferred.

It enrolls the following resources/properties in the CloudFormation docs:
- AWS::EC2::Volume
- AWS::EC2::VPCEndpoint
- AWS::ECR::Repository.EncryptionConfiguration.EncryptionType
- AWS::IAM::Role.Policies
- AWS::IAM::Role.RoleName

We cannot automatically match which descriptions should get replaced by the CloudFormation docs because the two have subtle differences (extra/different punctuation marks, CFN vs AWS CloudFormation, etc.)

Fixes #1572
Fixes #1550


provider/pkg/schema/gen.go
```diff
@@ -983,6 +994,11 @@ func (ctx *cfSchemaContext) updateDesc(refName, propName string, spec *jsschema.
 				}
 			}
 		}
+	} else if _, ok := forceDocumentationAugmentation[GetFullyQualifiedPropertyName(refName, propName)]; ok {
+		if desc, found := ctx.docs.GetPropertyDesc(refName, propName); found {
```
thomas11, line 998: Else: warn or report if augmentation was configured but not found?

thomas11 (commented):
Nice! A test would be good, though.

provider/pkg/schema/gen.go
flostadler: Yeah that's a good idea! I'm adding it to the report

thomas11 (approved):

t0yv0:
🚢

PR #1675, opened 2024-08-20
Nightly SDK regeneration

### LINKED ISSUE #1674: Workflow failure: nightly-sdk-generation
## Workflow Failure

[nightly-sdk-generation](https://github.com/pulumi/pulumi-aws-native/blob/master/.github/workflows/nightly-sdk-generation.yml) has failed. See the list of failures below:

- [2024-08-20T04:45:04.000Z](https://github.com/pulumi/pulumi-aws-native/actions/runs/10465399321)

flostadler:
The SDK had to be manually regenerated because an empty string was added to the CFN schema of the `AWS::IVS::Channel.Preset" enum:
```
"Preset" : {
      "description" : "Optional transcode preset for the channel. This is selectable only for ADVANCED_HD and ADVANCED_SD channel types. For those channel types, the default preset is HIGHER_BANDWIDTH_DELIVERY. For other channel types (BASIC and STANDARD), preset is the empty string (\"\").",
      "type" : "string",
      "enum" : [ "", "HIGHER_BANDWIDTH_DELIVERY", "CONSTRAINED_BANDWIDTH_DELIVERY" ]
    }
```

The [CloudFormation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ivs-channel.html#cfn-ivs-channel-preset) and [AWS CLI docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ivs/create-channel.html) mention that the allowed values are `HIGHER_BANDWIDTH_DELIVERY`, `CONSTRAINED_BANDWIDTH_DELIVERY`. But the empty string is also allowed and it's the default the IVS service sets internally (and returns) when a channel type is used that doesn't allow configuring a preset.

This change adds a special condition to the schema generator to create an enum value called `Empty` with an underlying value of `""`. We cannot omit this empty enum value as this would break SDKs that verify whether an enum value is valid (e.g. python).

Fixes #1674


provider/pkg/schema/gen.go
```diff
@@ -1431,6 +1431,12 @@ func (ctx *cfSchemaContext) genEnumType(enumName string, propSchema *jsschema.Sc
 			continue
 		}
 
+		// Special case for `ChannelPreset` enum in the `AWS::IVS::Channel` resource. The enum has an empty string value which
+		// gets returned by the service in certain default cases (for channel types (BASIC and STANDARD)).
+		if (typName == "ChannelPreset" && str == "") {
```
flostadler, line 1436: This is the actual change that makes the generation work

corymhall (approved):

Issue #1706, opened 2024-09-05
ReplaceOnChanges not set for nested properties


t0yv0:
### What happened?

It looks like there is some existing code around markCreateOnlyProperty that is incomplete with respect to nested properties. 

https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/schema/gen.go#L667

### Example

Consider AWS::LookoutMetrics::AnomalyDetector example.

AWS CF schema specifies: 

```
  "createOnlyProperties" : [ "/properties/AnomalyDetectorName", "/properties/MetricSource" ],
```

However this is not a top-level property. It's an object under definitions:

```
    "MetricSource" : {
      "type" : "object",
      "properties" : {
        "S3SourceConfig" : {
          "$ref" : "#/definitions/S3SourceConfig"
        },
        "RDSSourceConfig" : {
          "$ref" : "#/definitions/RDSSourceConfig"
        },
        "RedshiftSourceConfig" : {
          "$ref" : "#/definitions/RedshiftSourceConfig"
        },
        "CloudwatchConfig" : {
          "$ref" : "#/definitions/CloudwatchConfig"
        },
        "AppFlowConfig" : {
          "$ref" : "#/definitions/AppFlowConfig"
        }
      },
      "additionalProperties" : false
    },

```

Used as a nested declaration of another property:

```
    "MetricSet" : {
      "type" : "object",
      "properties" : {
        "MetricSource" : {
          "$ref" : "#/definitions/MetricSource"
        },

```

Currently `make local_generate` started failing catastrophically on encountering this. To workaround the problem, I will temporarily introduce a patch to ignore these failures, but we possibly need to revisit later on for completeness.

AFAIK replaceOnChanges property in the schema is of a documentation value at the moment and does not affect the actual behavior of the provider, but this may change.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Turns out I was mistaken, this is a warning and not a fatal error so it is simply a nice-to-have.

PR #1707, opened 2024-09-05
Fix nightly generation failure

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws-native/issues/1704

Review commit by commit.

First add more visible log fencing - without that I first incorrectly diagnosed the blocker as https://github.com/pulumi/pulumi-aws-native/issues/1706 where it was another issue repsonsible.

Second fix the blocker issue: instead of failing on dangling references, tolerate them.

### DIFF
--- a/provider/cmd/pulumi-gen-aws-native/main.go
+++ b/provider/cmd/pulumi-gen-aws-native/main.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -85,7 +86,7 @@ func main() {
 		}
 
 		if err := downloadCloudFormationDocs(docsUrl, filepath.Join(".", "aws-cloudformation-docs")); err != nil {
-			panic(fmt.Errorf("error download CloudFormation docs: %v", err))
+			fatalf("error download CloudFormation docs: %v", err)
 		}
 
 	case "discovery":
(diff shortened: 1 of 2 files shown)


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -96,69 +97,75 @@ func main() {
...
-		panic(fmt.Sprintf("Unrecognized language '%s'", operation))
+		fatalf("Unrecognized language '%s'", operation)
 	}
 }
 
+// Fail loud and clear.
+func fatalf(format string, a ...any) {
+	barrier := strings.Repeat("=", 110)
+	log.Fatalf(fmt.Sprintf("schema generation failed\n%s\n%s\n%s\n", barrier, fmt.Sprintf(format, a...), barrier))
+}
```
mjeffryes, line 167: Are we not concerned that jobs that used to depend on hard failure from this step will now pass with a silent failure in the logs?

mjeffryes (approved):

provider/cmd/pulumi-gen-aws-native/main.go
t0yv0: log.Fatalf is similar to panic in setting exit code 1.

t0yv0: I think it just adds a date. We could switch back to panic if that's better (it has stack trace). But these errors I think seem to be sufficiently annotated with wrappers that stack trace is not needed.

PR #1727, opened 2024-09-12
Auto generate AWS regions based on botocore source

### LINKED ISSUE #1726: Missing regions in Region enum
## Hello!

## Issue details
There are a number of AWS Regions that are not available in awsn.Region. For example, il-central-1. Without support of these values it is impossible to instantiate a custom provider in typescript, for example, as it results in a compilation error.

This issue also existed in the classic AWS provider, was reported in https://github.com/pulumi/pulumi-aws/issues/4402 and fixed in https://github.com/pulumi/pulumi-aws/pull/4404. It should be fixed in aws native as well.

flostadler:
Botocore is the heart of the AWS python SDK and the CLI. It has the most up-to-date information about regions and partitions.
We can use that information as part of our nightly SDK gen to auto generate the region enums and stay up-to-date with AWS regions.

fixes #1726


provider/cmd/pulumi-resource-aws-native/schema.json
```diff
@@ -57538,30 +57543,55 @@
                     "value": "ap-southeast-2"
                 },
                 {
-                    "name": "CaCentral",
```
flostadler, line 57541: Meh, seems like we accidentally forgot the `1` at the end when we created that enum.
I'm gonna manually add this one and deprecate it

Makefile
```diff
@@ -45,6 +46,17 @@ discovery:: update_submodules codegen
...
+	# Fetch the latest version of botocore
+	git -c 'versionsort.suffix=-' ls-remote --exit-code --refs --sort='version:refname' --tags https://github.com/boto/botocore '*.*.*' \
+		| tail -n 1 \
+		| cut -d '/' -f 3 \
+		> "$(METADATA_DIR)/.botocore.version"
+
+	# Extract the regions from botocore
+	cat $(METADATA_DIR)/.botocore.version | xargs -I{} curl -fsSL https://raw.githubusercontent.com/boto/botocore/{}/botocore/data/endpoints.json \
+		| jq '[.partitions[].regions | to_entries[] | { name: .key, description: .value.description}]' \
+		> "$(METADATA_DIR)/regions.json"
```
corymhall, line 58: Nice!

provider/pkg/schema/gen_test.go
```diff
@@ -487,3 +500,52 @@ func TestGatherPackage_docs_augmentation_unknownProperty(t *testing.T) {
...
+	})
+
+	assert.Contains(t, packageSpec.Types, "aws-native:index/Region:Region")
+	regionEnum := packageSpec.Types["aws-native:index/Region:Region"].Enum
+
+	assert.Equal(t, []schema.EnumValueSpec{
+		{Value: "cn-north-1", Name: "CnNorth1", Description: "China (Beijing)"},
+		{Value: "eu-isoe-west-1", Name: "EuIsoeWest1", Description: "EU ISOE West"},
+		{Value: "us-east-1", Name: "UsEast1", Description: "US East (N. Virginia)"},
+		{Value: "us-gov-west-1", Name: "UsGovWest1", Description: "AWS GovCloud (US-West)"},
```
corymhall, line 548: It's fun that we originally made the choice for the region enums to not match what we have in `pulumi-aws` 🙃

PR #1791, opened 2024-10-31
Fix cidr intrinsic

### LINKED ISSUE #593: aws-native.cidr function does not match CloudFormation's Fn::Cidr
## Hello!

## Issue details

I tried to use `aws-native.cidr` by assuming it corresponded to the CloudFormation function `Fn::Cidr`. However, it always failed immediately for more than one subnet. I dug into the implementation and found that it was a very different function. I assume this is an implementation oversight, but it might be a documentation issue.

The CloudFormation function [Fn::Cidr](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-cidr.html):
 - takes a larger subnet to divide into smaller subnets (`ipBlock`)
 - interprets `cidrBits` as the size of the resulting subnet (the inverse of the prefix length)

The `aws-native.cidr` function:
 - takes a base subnet to find siblings (`ipBlock`)
 - interprets `cidrBits` as the prefix length (the inverse of the size the resulting subnet)

Also the `aws-native.cidr` function seems to interpret [gocidr.NextSubnet](https://github.com/apparentlymart/go-cidr/blob/d07067359aee8ee9101c1a79a4f286cf0cea059f/cidr/cidr.go#L185-L199)'s "overflow" return value as an "ok" value, so it always immediately fails for more than one subnet with:
```
    error: Error: invocation of aws-native:index:cidr returned an error: could not create 3 subnets
```

### Affected area/feature

Assuming `cidr` function is intended to correspond to the CloudFormation version, then I believe the required change is in [provider/pkg/provider/provider_intrinsics.go](https://github.com/pulumi/pulumi-aws-native/blob/dfb123ac313acbea0068b697df2c63945640263e/provider/pkg/provider/provider_intrinsics.go#L45-L90).

### LINKED ISSUE #1566: cidr{Output} function doesn't work?
### What happened?

```
const cidr = aws.cidrOutput({
    ipBlock: "2a05:d024:d::/56",
    cidrBits: 4,
    count: 1,
})
```

I expected this to create a `/60` CIDR. Instead ot just returns the string "2a05:d024:d::/56"

### Example

```
const cidr = aws.cidrOutput({
    ipBlock: "2a05:d024:d::/56",
    cidrBits: 4,
    count: 1,
})
```

### Output of `pulumi about`

```
CLI          
Version      3.90.1
Go Version   go1.21.9
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.37.1
aws-native  0.99.0
nodejs      unknown
```

### Additional context

Same problem occurs with IPv$

### Contributing

corymhall:
The `cidr` intrinsic function has been broken. This PR fixes the cidr function and adds some tests. Took some inspiration from [OpenTofu cidrsubnets](https://github.com/opentofu/opentofu/blob/9d842aa920e16cc3275f4fe54240634fff5146dc/internal/lang/funcs/cidr.go#L123)

fixes #1566, fixes #593

### DIFF
--- a/provider/pkg/provider/provider_intrinsics.go
+++ b/provider/pkg/provider/provider_intrinsics.go
@@ -87,7 +87,10 @@ func cidr(inputs resource.PropertyMap) (resource.PropertyMap, error) {
 		return nil, fmt.Errorf("cidrBits %d would extend prefix to %d bits, which is too long for an %s address", int(cidrBits.NumberValue()), prefixLen, protocol)
 	}
 
-	current, _ := gocidr.PreviousSubnet(network, prefixLen)
+	current, ok := gocidr.PreviousSubnet(network, prefixLen)
+	if ok || !network.Contains(current.IP) {
+		return nil, fmt.Errorf("not enough remaining address space for a subnet with a prefix of %d bits after %s", len(subnets), current.String())
+	}
 	for i := 0; i < len(subnets); i++ {
 		subnet, ok := gocidr.NextSubnet(current, prefixLen)
 		if ok || !network.Contains(subnet.IP) {
--- a/provider/pkg/provider/provider_intrinsics.go
+++ b/provider/pkg/provider/provider_intrinsics.go
@@ -88,7 +88,8 @@ func cidr(inputs resource.PropertyMap) (resource.PropertyMap, error) {
 	}
 
 	current, ok := gocidr.PreviousSubnet(network, prefixLen)
-	if ok || !network.Contains(current.IP) {
+	// ok is true if we have rolled over (which we don't want)
+	if ok {
 		return nil, fmt.Errorf("not enough remaining address space for a subnet with a prefix of %d bits after %s", len(subnets), current.String())
 	}
 	for i := 0; i < len(subnets); i++ {
--- a/provider/pkg/provider/provider_intrinsics.go
+++ b/provider/pkg/provider/provider_intrinsics.go
@@ -42,7 +42,7 @@ func (p *cfnProvider) getAZs(ctx context.Context, inputs resource.PropertyMap) (
 	}), nil
 }
 
-func (p *cfnProvider) cidr(ctx context.Context, inputs resource.PropertyMap) (resource.PropertyMap, error) {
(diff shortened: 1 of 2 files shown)


provider/pkg/provider/provider_intrinsics.go
```diff
@@ -73,22 +73,39 @@ func (p *cfnProvider) cidr(ctx context.Context, inputs resource.PropertyMap) (re
...
+		}
+		return nil, fmt.Errorf("cidrBits %d would extend prefix to %d bits, which is too long for an %s address", int(cidrBits.NumberValue()), prefixLen, protocol)
+	}
 
-	prefixLen := int(cidrBits.NumberValue())
-	for i := 1; i < len(subnets)-1; i++ {
-		subnet, ok := gocidr.NextSubnet(network, prefixLen)
-		if !ok {
-			return nil, fmt.Errorf("could not create %d subnets", len(subnets))
+	current, _ := gocidr.PreviousSubnet(network, prefixLen)
```
t0yv0, line 90: What's the underscore here? Should we `contract.Assertf(ok, "PreviousSubnet lookup failed")` or inspect it?

provider/pkg/provider/provider_intrinsics_test.go
```diff
@@ -0,0 +1,88 @@
...
+		assert.ErrorContains(t, err, "cidrBits 80 would extend prefix to 136 bits, which is too long for an IPv6 address")
+	})
+
+	t.Run("ipv6 not enough space", func(t *testing.T) {
+		_, err := cidr(resource.NewPropertyMapFromMap(map[string]interface{}{
+			"ipBlock":  "2600:1f16:44e:3e00::/56",
+			"count":    3,
+			"cidrBits": 1,
+		}))
+		assert.ErrorContains(t, err, "not enough remaining address space for a subnet with a prefix of 3 bits after 2600:1f16:44e:3e80::/57")
```
t0yv0, line 85: Thank you! Nice!!

t0yv0 (approved):

PR #1793, opened 2024-11-01
chore: clarify DiffConfig code

t0yv0:
This change does not affect the code behavior but makes it clear that AWS Native never cascade-replaces resources in DiffConfig.


flostadler (approved):
LGTM

Issue #16, opened 2021-03-10
Make Diff and Read "stable"


mikhailshilkov:
Currently, Diff always returns an update because the attributes (read-only values) are missing from the inputs. We should ignore attributes when calculating diffs.

Similarly, `Read` should exclude attributes when calculating inputs (there a TODO in the code for that).

Likely, we need to introduce the `__inputs` pattern in order to have a reliable way to compare old inputs with new inputs.

PR #1795, opened 2024-11-04
Refactor Custom Resources to allow configuring timeouts and secrets

flostadler:
Previously it was assumed that the custom resources do not concern themselves with secretness or handle timeouts.
But this is necessary to support CloudFormation based Custom Resources.
Those can contain secret outputs and support configuring timeouts.

This change modifies the `CustomResource` interface to accommodate those requirements.
In detail this means that the Create/Update/Read lifecycle methods now returns a `PropertyMap` instead of a generic `map[string]interface{}` and take a timeout
parameter.

For reviewing this change, I'd first recommend having a look at the changes to the `CustomResource` interface in `provider/pkg/resources/custom.go` and then double check the refactoring changes resulting in that.

This change also introduces tests for the provider's CRUD lifecycle. As part of doing that, I added mocks using `uber/gomock`.

Relates to https://github.com/pulumi/pulumi-cdk/issues/109


flostadler:
The test coverage of the touched files is mostly through integration tests. I'll try to add some more tests to cover parts of this with units as well

corymhall (commented):
Does this change have any user facing impact, or is it purely internal?

flostadler:
> Does this change have any user facing impact, or is it purely internal?

This is purely internal

provider/pkg/provider/provider.go
```diff
@@ -830,11 +830,11 @@ func (p *cfnProvider) Create(ctx context.Context, req *pulumirpc.CreateRequest)
 
 	resourceToken := string(urn.Type())
 	var id *string
-	var outputs map[string]interface{}
+	var outputs resource.PropertyMap
 	var createErr error
-	var payload map[string]interface{}
+	timeout := time.Duration(req.GetTimeout()) * time.Second
```
t0yv0, line 835: Unlike bridged providers, guessing there is no need to merge with schema-derived timeouts here?

Also, double-checking, this will be `0` if req.GetTimeout() is unset?

provider/pkg/provider/provider.go
```diff
@@ -1173,22 +1178,6 @@ func (p *cfnProvider) Cancel(context.Context, *pbempty.Empty) (*pbempty.Empty, e
 	return &pbempty.Empty{}, nil
 }
 
-// checkpointObject puts inputs in the `__inputs` field of the state.
-func checkpointObject(inputs resource.PropertyMap, outputs map[string]interface{}) resource.PropertyMap {
-	object := resource.NewPropertyMapFromMap(outputs)
-	object["__inputs"] = resource.MakeSecret(resource.NewObjectProperty(inputs))
```
t0yv0, line 1179: This was a fairly interesting quirk.

provider/pkg/provider/provider.go
flostadler: No, there's no notion of passing timeouts to the CloudControl based resources (yet).

At the current time no resource is using timeouts yet. This is just prep for the next changes related to CFN CustomResources

flostadler: And yes, this will be `0` if no timeouts were specified

provider/pkg/resources/checkpoint.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2024, Pulumi Corporation.
+
+package resources
+
+import (
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+// CheckpointObject puts inputs in the `__inputs` field of the state.
```
t0yv0, line 9: Out of scope here but I am curious now do we understand why it's needed? Throwing a sentence on why we're doing this could be interesting.

provider/pkg/resources/custom.go
```diff
@@ -12,11 +13,11 @@ type CustomResource interface {
 	// Check validates and transforms the inputs of the resource.
 	Check(ctx context.Context, urn resource.URN, randomSeed []byte, inputs, state resource.PropertyMap, defaultTags map[string]string) (resource.PropertyMap, []ValidationFailure, error)
 	// Create creates a new resource in the cloud provider and returns its unique identifier and outputs.
-	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs map[string]any, err error)
+	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap, timeout time.Duration) (identifier *string, outputs resource.PropertyMap, err error)
 	// Read returns the outputs and the updated inputs of the resource.
-	Read(ctx context.Context, urn resource.URN, id string, oldInputs, oldOutputs resource.PropertyMap) (outputs map[string]any, inputs resource.PropertyMap, exists bool, err error)
+	Read(ctx context.Context, urn resource.URN, id string, oldInputs, oldOutputs resource.PropertyMap) (outputs resource.PropertyMap, inputs resource.PropertyMap, exists bool, err error)
```
t0yv0, line 18: Love a little more type-safety here.

provider/pkg/resources/custom.go
```diff
@@ -12,11 +13,11 @@ type CustomResource interface {
 	// Check validates and transforms the inputs of the resource.
 	Check(ctx context.Context, urn resource.URN, randomSeed []byte, inputs, state resource.PropertyMap, defaultTags map[string]string) (resource.PropertyMap, []ValidationFailure, error)
 	// Create creates a new resource in the cloud provider and returns its unique identifier and outputs.
-	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap) (identifier *string, outputs map[string]any, err error)
+	Create(ctx context.Context, urn resource.URN, inputs resource.PropertyMap, timeout time.Duration) (identifier *string, outputs resource.PropertyMap, err error)
```
t0yv0, line 16: This seems like a breaking change but we are expecting this interface to be internal? .. We're on `v1.5.0`. Nothing consumes it in other codebases? Should we work harder here as in introducing a side-by-side interface?

t0yv0 (approved):

provider/pkg/resources/custom.go
flostadler: Yeah, this interface is internal. There's no way somebody could use this interface and plug it into aws-native unless they fork aws-native. It's used internally to wrap arbitrary custom resources: https://github.com/pulumi/pulumi-aws-native/blob/0166a6357100a309d815a53cc4ac986c3fa96aab/provider/pkg/provider/provider.go#L509-L511

flostadler:
> The test coverage of the touched files is mostly through integration tests. I'll try to add some more tests to cover parts of this with units as well

I added unit tests for the provider's CRUD operations. Coverage wise we should be good now and I managed to uncover a bug: https://github.com/pulumi/pulumi-aws-native/issues/1796

provider/pkg/resources/checkpoint.go
flostadler: This was done to make diff/read more robust (https://github.com/pulumi/pulumi-aws-native/pull/17) and solve this issue: https://github.com/pulumi/pulumi-aws-native/issues/16.

In detail, having access to the old inputs during Read.
I'll put this in the docs in a separate PR

corymhall (approved):

PR #1797, opened 2024-11-04
Fix cidr function

corymhall:
In #1791 we \"fixed\" the cidr function so that it would generate subnets
correctly. We did not fix it correctly though. This cidr function needs
to match the logic that the CloudFormation Fn::Cidr function has.

Critically the `cidrBits` logic was incorrect. It should be the inverse
of the subnet mask that you want to have for the subnets.

For example a `cidrBits` of 5 means that I want subnets with a mask of
`/27` (32-5=27).

```ts
cidr({
  ipBlock: '192.168.0.0/24',
  count: 6,
  cidrBits: 5,
});
```

### DIFF
--- a/examples/cidr-ts/index.ts
+++ b/examples/cidr-ts/index.ts
@@ -1,0 +1,48 @@
+// Copyright 2016-2021, Pulumi Corporation.
+
+import * as pulumi from '@pulumi/pulumi';
+import * as aws from "@pulumi/aws-native";
+
+const cidrBlock = '192.168.0.0/24';
+const vpc = new aws.ec2.Vpc('vpc', {
+    cidrBlock,
+});
+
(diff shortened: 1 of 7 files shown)


examples/cidr-ts/index.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright 2016-2021, Pulumi Corporation.
```
flostadler, line 1:
```suggestion
// Copyright 2016-2024, Pulumi Corporation.
```

flostadler (approved):
Nice!

t0yv0 (approved):
Sounds like it's not an implementation issue but we mis-understood the spec and didn't cross-tests against CloudFormation expected values. Great fix!

PR #1801, opened 2024-11-05
Extend CustomResource interface to pass state for Update and Delete

flostadler:
Previously the Update and Delete methods for Custom Resources
did not require access to the current state, but this is a
common requirement for updating and deleting resources.

This is also needed for supporting CloudFormation Custom Resources
as those require access to state in order to complete updates
and deletions.

relates to https://github.com/pulumi/pulumi-cdk/issues/109

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -1062,7 +1062,7 @@ func (p *cfnProvider) Update(ctx context.Context, req *pulumirpc.UpdateRequest) 
 	if customResource, ok := p.customResources[resourceToken]; ok {
 		timeout := time.Duration(req.GetTimeout()) * time.Second
 		// Custom resource
-		outputs, err = customResource.Update(ctx, urn, id, newInputs, oldInputs, timeout)
+		outputs, err = customResource.Update(ctx, urn, id, newInputs, oldInputs, oldState, timeout)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


provider/pkg/provider/provider.go
```diff
@@ -1129,8 +1129,13 @@ func (p *cfnProvider) Delete(ctx context.Context, req *pulumirpc.DeleteRequest)
 			return nil, errors.Wrapf(err, "failed to parse inputs for update")
 		}
 
+		// Retrieve the state.
+		state, err := plugin.UnmarshalProperties(req.GetProperties(), plugin.MarshalOptions{
+			Label: fmt.Sprintf("%s.state", label), KeepUnknowns: true, SkipNulls: true, KeepSecrets: true,
```
t0yv0, line 1134: Curious, preferring SkipNulls intentionally? 

Also, KeepUnknowns does not matter as there are no unknowns in state I think.

t0yv0 (approved):

t0yv0:
Update must have access to state . This looks like it is fixing an important omission. 🚢

provider/pkg/provider/provider.go
flostadler: This is to stay consistent with how we're reading the state in other life-cycle methods (e.g. Read: https://github.com/pulumi/pulumi-aws-native/blob/2962615c932c7ea3d791f8500acb2bf285d69251/provider/pkg/provider/provider.go#L919C2-L921C4)

PR #1805, opened 2024-11-07
Add AWS SDK clients for CFN Custom Resource support

flostadler:
This change adds the necessary AWS SDK clients (S3 and Lambda) for supporting CFN Custom Resources in aws-native.

Relates to https://github.com/pulumi/pulumi-cdk/issues/109

PR #1806, opened 2024-11-07
Add implementation for CloudFormation Custom Resource Emulator

flostadler:
This PR adds support for [CloudFormation Custom Resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html) to the aws-native provider. It implements an emulator that enables Pulumi programs to interact with Lambda-backed CloudFormation Custom Resources.

A CloudFormation custom resource is essentially an extension point to run arbitrary code as part of the CloudFormation lifecycle. It is similar in concept to the [Pulumi Command Provider](https://www.pulumi.com/registry/packages/command/), the difference being that CloudFormation CustomResources are executed in the Cloud; either through Lambda or SNS.

For the first implementation we decided to limit the scope to Lambda backed Custom Resources, because the SNS variants are not widely used.

## Custom Resource Protocol
The implementation follows the CloudFormation Custom Resource protocol. I derived the necessary parts by combining information from the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref.html), [CDKs CustomResource Framework](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/custom-resource-handlers/lib/custom-resources-framework) and trial&error.

Notable aspects of that protocol are:
- primitive properties need to be string encoded when sending them to Custom Resource handlers. This includes deeply nested properties: https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1037
- The Lambda Function is invoked asynchronously. Lambda will retry the execution if the function fails unexpectedly (e.g. unhandled exception).
- Due to the async invocation, the response is not returned from the Lambda Function, instead it's sent to a `ResponseURL` that needs to be included in the request payload.
    - Similarly to CloudFormation, we decided to implement this using S3 Buckets and presigned URLs.

### Custom Resource Lifecycle
```mermaid
sequenceDiagram
    participant A as aws-native
    participant S3 as S3 Bucket
    participant L as Lambda
    
    %% Create Flow
    Note over A,L: Create Operation
    A->>S3: Generate presigned URL
    A->>L: Invoke with CREATE event
    activate L
    loop Until response found or timeout
        A->>S3: Poll for response
        L-->>S3: Upload response
    end
    deactivate L
    A->>S3: Fetch response
    alt Success
        A->>A: Store PhysicalId & outputs
    else Failure
        A->>A: Return error
    end

    %% Update Flow
    Note over A,L: Update Operation
    A->>S3: Generate presigned URL
    A->>L: Invoke with UPDATE event
    activate L
    loop Until response found or timeout
        A->>S3: Poll for response
        L-->>S3: Upload response
    end
    deactivate L
    A->>S3: Fetch response
    alt Success
        A->>A: Check PhysicalId
        alt ID Changed
            A->>S3: Generate presigned URL for cleanup
            A->>L: Invoke with DELETE event for old resource
            activate L
            loop Until cleanup response found or timeout
                A->>S3: Poll for cleanup response
                L-->>S3: Upload cleanup response
            end
            deactivate L
            A->>S3: Fetch cleanup response
        end
    else Failure
        A->>A: Return error
    end

    %% Delete Flow
    Note over A,L: Delete Operation
    A->>S3: Generate presigned URL
    A->>L: Invoke with DELETE event
    activate L
    loop Until response found or timeout
        A->>S3: Poll for response
        L-->>S3: Upload response
    end
    deactivate L
    A->>S3: Fetch response
    alt Success
        A->>A: Return success
    else Failure
        A->>A: Return error
    end
```

## Reviewer Notes

Key areas to review:
1. Error handling in the response collection mechanism
2. Timeout management, especially for the `Update` lifecycle
3. Documentation completeness and accuracy

Exposing this resource and schematizing it is part of this PR https://github.com/pulumi/pulumi-aws-native/pull/1807.
Automatically cleaning up the response objects is not included in this PR in order to keep its size manageable. Implementing this is tracked here: https://github.com/pulumi/pulumi-aws-native/issues/1813.

Please pay special attention to:
- S3 response collection mechanism security
- State management during updates
- Cleanup handling when physical resource IDs change

## Testing
- Unit tests including error handling tests for various failure scenarios
- Integration tests with actual Lambda functions are added in this stacked PR: https://github.com/pulumi/pulumi-aws-native/pull/1807

## Related Issues
- https://github.com/pulumi/pulumi-cdk/issues/109
- https://github.com/pulumi/pulumi-aws-native/issues/1812
- https://github.com/pulumi/pulumi-aws-native/issues/1813

### DIFF
--- a/provider/pkg/resources/cfn_custom_resource.go
+++ b/provider/pkg/resources/cfn_custom_resource.go
@@ -250,7 +250,7 @@ func (c *cfnCustomResource) Check(ctx context.Context, urn urn.URN, randomSeed [
 	}
 
 	if typedInputs.StackID == nil {
-		// if the stack ID is not provided, we use the project ID as the stack ID
+		// if the stack ID is not provided, we use the pulumi stack ID as the stack ID
 		inputs[resource.PropertyKey("stackId")] = resource.NewStringProperty(urn.Stack().String())
 	}
 
--- a/provider/pkg/resources/cfn_custom_resource.go
+++ b/provider/pkg/resources/cfn_custom_resource.go
@@ -53,7 +53,7 @@ func NewCfnCustomResource(providerName string, s3Client client.S3Client, lambdaC
 		providerName: providerName,
 		s3Client:     s3Client,
 		lambdaClient: lambdaClient,
-		clock: &realClock{},
+		clock:        &realClock{},
 	}
 }
 
@@ -138,7 +138,7 @@ func CfnCustomResourceSpec() pschema.ResourceSpec {
 			"bucketName": {
 				Description: "The name of the S3 bucket to use for storing the response from the Custom Resource.\n\n" +
 					"The IAM principal configured for the provider must have `s3:PutObject`, `s3:HeadObject` and `s3:GetObject` permissions on this bucket.",
-				TypeSpec:    pschema.TypeSpec{Type: "string"},
+				TypeSpec: pschema.TypeSpec{Type: "string"},
 			},
 			"bucketKeyPrefix": {
 				Description: "The prefix to use for the bucket key when storing the response from the Custom Resource provider.",
@@ -526,6 +526,10 @@ func (i CfnCustomResourceInputs) makeOutputs(inputs resource.PropertyMap, respon
 }
 
 func sanitizeCustomResourceResponse(event *cfn.Event, response *cfn.Response) *cfn.Response {
+	if response == nil || response.PhysicalResourceID == "" {
+		return response
+	}
+
 	// ensure PhysicalResourceID is set. For Create requests we fall back to the RequestID,
 	// for Update and Delete requests we fall back to the PhysicalResourceID from state
 	if response.PhysicalResourceID == "" && (event.RequestType == cfn.RequestDelete || event.RequestType == cfn.RequestUpdate) {
--- a/provider/pkg/resources/cfn_custom_resource.go
+++ b/provider/pkg/resources/cfn_custom_resource.go
@@ -1,0 +1,538 @@
+package resources
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/aws/aws-lambda-go/cfn"
+	"github.com/golang/glog"
+	"github.com/google/uuid"
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/client"
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/naming"
+	"github.com/pulumi/pulumi-go-provider/resourcex"
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/urn"
+)
+
+// This is the default timeout for custom resource operations in CloudFormation
+const DefaultCustomResourceTimeout = 60 * time.Minute
+
+var lambdaFunctionArnRegex = regexp.MustCompile(`^arn:[^:]+:lambda:[^:]+:[^:]+:function:[a-zA-Z0-9-_]+(:[a-zA-Z0-9-_]+)?$`)
+
+type Clock interface {
+	Now() time.Time
+	Since(time.Time) time.Duration
+}
+
+type realClock struct{}
+
+func (realClock) Now() time.Time {
+	return time.Now()
+}
+
+func (realClock) Since(t time.Time) time.Duration {
+	return time.Since(t)
+}
+
+type cfnCustomResource struct {
+	providerName string
+	lambdaClient client.LambdaClient
+	s3Client     client.S3Client
+	clock        Clock
+}
+
+// Check CfnCustomResource implements CustomResource
+var _ CustomResource = (*cfnCustomResource)(nil)
+
+func NewCfnCustomResource(providerName string, s3Client client.S3Client, lambdaClient client.LambdaClient) *cfnCustomResource {
+	return &cfnCustomResource{
+		providerName: providerName,
+		s3Client:     s3Client,
+		lambdaClient: lambdaClient,
+		clock: &realClock{},
+	}
+}
+
+type CfnCustomResourceInputs struct {
+	// The name of the S3 bucket to use for storing the response from the custom resource
+	BucketName string
+	// The prefix to use for the bucket key when storing the response from the custom resource
+	BucketKeyPrefix string
+	// The service token, such as a Lambda function ARN. The service token must be in the same Region as this resource
+	ServiceToken string
+	// The custom resource properties to pass as an input to the Lambda function
+	CustomResourceProperties map[string]interface{}
+	// The CloudFormation type of the custom resource (e.g. "Custom::MyCustomResource")
+	ResourceType string
+	// A stand-in value for the CloudFormation stack ID required by the custom resource. If not provided, the project ID is used.
+	StackID *string
+}
+
+type CfnCustomResourceState struct {
+	// The physical resource ID of the custom resource
+	PhysicalResourceID string `json:"physicalResourceId"`
+	// The response data returned by invoking the custom resource lambda
+	Data map[string]interface{} `json:"data"`
+	// The stand-in value for the CloudFormation stack ID required by the custom resource
+	StackID string `json:"stackId"`
+	// The service token, such as a Lambda function ARN. The service token must be in the same Region as this resource
+	ServiceToken string `json:"serviceToken"`
+	// The name of the S3 bucket to use for storing the response from the custom resource
+	Bucket string `json:"bucket"`
+	// The CloudFormation type of the custom resource (e.g. "Custom::MyCustomResource")
+	ResourceType string `json:"resourceType"`
+	// Whether the response data contains sensitive information that should be marked as secret and not logged
+	NoEcho bool `json:"noEcho"`
+}
+
+func (s CfnCustomResourceState) ToPropertyMap() resource.PropertyMap {
+	return resource.NewPropertyMap(s)
+}
+
+func CfnCustomResourceSpec() pschema.ResourceSpec {
+	return pschema.ResourceSpec{
+		ObjectTypeSpec: pschema.ObjectTypeSpec{
+			Description: "TODO",
+			Properties: map[string]pschema.PropertySpec{
+				"physicalResourceId": {
+					Description: "The name or unique identifier that corresponds to the `PhysicalResourceId` included in the Custom Resource response. If no `PhysicalResourceId` is provided in the response, a random ID will be generated.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
+				"data": {
+					Description: "The response data returned by invoking the Custom Resource.",
+					TypeSpec: pschema.TypeSpec{
+						Type: "object",
+						AdditionalProperties: &pschema.TypeSpec{
+							Ref: "pulumi.json#/Any",
+						},
+					},
+				},
+				"stackId": {
+					Description: "A stand-in value for the CloudFormation stack ID.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
+				"serviceToken": {
+					Description: "The service token, such as a Lambda function ARN, that is invoked when the resource is created, updated, or deleted.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
+				"bucket": {
+					Description: "The name of the S3 bucket to use for storing the response from the Custom Resource.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
+				"resourceType": {
+					Description: "The CloudFormation type of the Custom Resource provider. For example, `Custom::MyCustomResource`.",
+					TypeSpec:    pschema.TypeSpec{Type: "string"},
+				},
+				"noEcho": {
+					Description: "Whether the response data contains sensitive information that should be marked as secret and not logged.",
+					TypeSpec:    pschema.TypeSpec{Type: "boolean"},
+				},
+			},
+			Required: []string{"physicalResourceId", "data", "stackId", "serviceToken", "bucket", "resourceType", "noEcho"},
+		},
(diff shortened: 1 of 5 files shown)


flostadler:

- #1805
    - #1806 ◀
        - #1807

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+// 2. Constructs a CloudFormation UPDATE event with:
+//    - Existing PhysicalResourceId
+//    - Old and new ResourceProperties
+//    - All standard CloudFormation fields
+// 3. Invokes Lambda and collects response
+// 4. If PhysicalResourceId changes:
+//    - Initiates cleanup of old resource
+//    - Sends DELETE event for old PhysicalResourceId
+// 5. Returns updated properties and new PhysicalResourceId
+func (c *cfnCustomResource) Update(ctx context.Context, urn urn.URN, id string, inputs, oldInputs, state resource.PropertyMap, timeout time.Duration) (resource.PropertyMap, error) {
```
corymhall, line 462: Is there a way for us to get the resource options here, specifically the `retainOnDelete` option? I know there were times that I would set the `retainOnDelete` to `true` for custom resources because they didn't handle delete well.

corymhall (approved):
Just a couple of minor comments, otherwise this is awesome! Great work!

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+	var typedInputs CfnCustomResourceInputs
+	_, err := resourcex.Unmarshal(&typedInputs, inputs, resourcex.UnmarshalOptions{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal inputs: %w", err)
+	}
+
+	var failures []ValidationFailure
+
+	if !lambdaFunctionArnRegex.MatchString(typedInputs.ServiceToken) {
+		// TODO create a GitHub issue for this and link it
```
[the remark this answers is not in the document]
flostadler: Thanks! Created the ticket but forgot removing the todo 🤦

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+		failures = append(failures, ValidationFailure{
+			Path:   "serviceToken",
+			Reason: "serviceToken must be a valid Lambda function ARN.",
+		})
+	}
+
+	if typedInputs.StackID == nil {
+		// if the stack ID is not provided, we use the project ID as the stack ID
```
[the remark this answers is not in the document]
flostadler: Yeah you're right. Changed the code, but not the comment..

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+
+	requestID := uuid.New().String()
+	bucketKey := fmt.Sprintf("%s/%s.json", invokeData.bucketKeyPrefix, requestID)
+
+	responseUrl, err := c.s3Client.PresignPutObject(ctx, invokeData.bucket, bucketKey, timeout)
+	if err != nil {
+		// presigning is not an API call, it should not fail unless there's issues with the SDK or crypto libs
+		return nil, fmt.Errorf("failed to generate response URL: %w", err)
+	}
+	glog.V(9).Infof("%s created presigned response URL %q for s3://%s/%s", invokeData.loggingLabel, responseUrl, invokeData.bucket, bucketKey)
```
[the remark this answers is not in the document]
flostadler: Uh, that's a good one! I'll remove it

provider/pkg/resources/cfn_custom_resource.go
flostadler: Sadly, not that I'm aware of: https://pulumi-developer-docs.readthedocs.io/latest/docs/_generated/proto/provider.html#pulumirpc-updaterequest

t0yv0: Only MLCs have access to retainOnDelete it seems:
https://github.com/pulumi/pulumi/blob/master/proto/pulumi/provider.proto#L473

For CustomResources that are traditional I don't think the engine ever  cooperates with the provider on this. The engine handles it all engine-side.

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+	}
+
+	outputs := newTypedInputs.makeOutputs(inputs, response)
+	return outputs, nil
+}
+
+// Read returns the current inputs and outputs of the custom resource because CFN custom resources do not store state.
+// They are just a stateless wrapper around a Lambda function or SNS topic.
+func (c *cfnCustomResource) Read(ctx context.Context, urn urn.URN, id string, oldInputs resource.PropertyMap, oldState resource.PropertyMap) (resource.PropertyMap, resource.PropertyMap, bool, error) {
+	if len(oldState) == 0 {
```
t0yv0, line 565: Comment here that len(oldState) is trying to guess if we are in a Read-for-import. 
The developer docs are improving but not quite there yet but are being improved (cc @lunaris )

https://pulumi-developer-docs.readthedocs.io/latest/developer-docs/providers/implementers-guide.html#read

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+					"The IAM principal configured for the provider must have `s3:PutObject`, `s3:HeadObject` and `s3:GetObject` permissions on this bucket.",
+				TypeSpec: pschema.TypeSpec{Type: "string"},
+			},
+			"bucketKeyPrefix": {
+				Description: "The prefix to use for the bucket key when storing the response from the Custom Resource provider.",
+				TypeSpec:    pschema.TypeSpec{Type: "string"},
+			},
+			"serviceToken": {
+				Description: "The service token to use for the Custom Resource. The service token is invoked when the resource is created, updated, or deleted.\n" +
+					"This can be a Lambda Function ARN with optional version or alias identifiers.\n\n" +
```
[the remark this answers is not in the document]
flostadler: Lambda function ARNs (including alias or version) are a very common concept in AWS. I don't think we should go into too much detail about it here.

This is also not a secret, ARNs are just unique IDs

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+}
+
+func (s CfnCustomResourceState) ToPropertyMap() resource.PropertyMap {
+	return resource.NewPropertyMap(s)
+}
+
+func CfnCustomResourceSpec() pschema.ResourceSpec {
+	return pschema.ResourceSpec{
+		ObjectTypeSpec: pschema.ObjectTypeSpec{
+			Description: "TODO: pulumi/pulumi-cdk#109",
```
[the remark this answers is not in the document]
flostadler: This is changed in the next stacked PR. Just needed a quick placeholder :)

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+)
+
+// This is the default timeout for custom resource operations in CloudFormation
+const DefaultCustomResourceTimeout = 60 * time.Minute
+
+var lambdaFunctionArnRegex = regexp.MustCompile(`^arn:[^:]+:lambda:[^:]+:[^:]+:function:[a-zA-Z0-9-_]+(:[a-zA-Z0-9-_]+)?$`)
+
+type Clock interface {
+	Now() time.Time
+	Since(time.Time) time.Duration
```
t0yv0, line 28: Nit: Go likes minimal interfaces so you only really need Now(), as Since can be computed from Now in a `func`.

flostadler: Yes, but that would clutter the implementation. The purpose of this interface is to make the code testable. I'd rather just pass through to the stdlib than add custom code

flostadler: Mhm, I guess this is a bit of a moot point from my side given that you can subtract times and get the duration that way

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+	timeout         time.Duration
+	loggingLabel    string
+	serviceToken    string
+}
+
+// Check validates the inputs of the resource and applies default values if necessary.
+// It returns the inputs, validation failures, and an error if the inputs cannot be unmarshalled.
+func (c *cfnCustomResource) Check(ctx context.Context, urn urn.URN, randomSeed []byte, inputs resource.PropertyMap, state resource.PropertyMap, defaultTags map[string]string) (resource.PropertyMap, []ValidationFailure, error) {
+	var typedInputs CfnCustomResourceInputs
+	_, err := resourcex.Unmarshal(&typedInputs, inputs, resourcex.UnmarshalOptions{})
```
t0yv0, line 237: I do not fully trust this function, and reading the docs quickly is not very reassuring some of the concerns I have. 

I think we might need to think about this a bit (probably out of scope of this PR but before being fully confident).

What happens with unknowns, dependencies and secrets?

For secrets it's a design question - custom resources do not support them right? So we need to support secrets through the black-box transformation I think?

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+//    - ResourceType from input
+//    - LogicalResourceId from Pulumi URN
+//    - Custom properties from input
+// 3. Invokes the Lambda function asynchronously
+// 4. Waits for response in S3 bucket
+// 5. Processes response:
+//    - On success: Returns PhysicalResourceId and properties
+//    - On failure: Returns error with reason
+//    - Handles `NoEcho` for sensitive data
+func (c *cfnCustomResource) Create(ctx context.Context, urn urn.URN, inputs resource.PropertyMap, timeout time.Duration) (*string, resource.PropertyMap, error) {
```
t0yv0, line 296: I was wondering if this is ever called in Preview, but looks like the provider is SupportsPreview: false so we're good. It never is! No unknowns then.

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -0,0 +1,681 @@
...
+	bucket          string
+	bucketKeyPrefix string
+	timeout         time.Duration
+	loggingLabel    string
+	serviceToken    string
+}
+
+// Check validates the inputs of the resource and applies default values if necessary.
+// It returns the inputs, validation failures, and an error if the inputs cannot be unmarshalled.
+func (c *cfnCustomResource) Check(ctx context.Context, urn urn.URN, randomSeed []byte, inputs resource.PropertyMap, state resource.PropertyMap, defaultTags map[string]string) (resource.PropertyMap, []ValidationFailure, error) {
```
t0yv0, line 235: I *think* this might still receive unknowns but I'm not 100% sure.

provider/pkg/resources/cfn_custom_resource_test.go
```diff
@@ -0,0 +1,1540 @@
...
+					"nestedProp1": "nestedValue1",
+					"nestedProp2": 42,
+				},
+			},
+			customResourceInputs: map[string]interface{}{
+				"key": "value",
+			},
+		},
+		{
+			name:   "SecretResponse",
```
t0yv0, line 191: Would it be possible to see here what is returned to the engine from Create in this test case?

Perhaps with `hexops/autogold` if it's too much trouble to write by hand.

flostadler: We're asserting this with the assertion in line 348. In case `noEcho` is set to true, we expect the `data` prop to be marked as a secret.

provider/pkg/resources/cfn_custom_resource_test.go
```diff
@@ -0,0 +1,1540 @@
...
+	t.Parallel()
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockLambdaClient := client.NewMockLambdaClient(ctrl)
+	mockS3Client := client.NewMockS3Client(ctrl)
+
+	expectedError := fmt.Errorf("failed to invoke lambda")
+	stackID := "stack-id"
+	serviceToken := "arn:aws:lambda:us-west-2:123456789012:function:my-function"
```
t0yv0, line 1424: Here's the serviceToken example, I see. Indeed it looks like a lambda ARN. Interesting.

t0yv0 (approved):
🚢

provider/pkg/resources/cfn_custom_resource_test.go
t0yv0: Ah ok!

provider/pkg/resources/cfn_custom_resource.go
t0yv0: Yeah it's really a small nit, feel free to disregard.

provider/pkg/resources/cfn_custom_resource.go
flostadler: We're already quite heavily using the `resourcex.Decode` and `resourcex.Unmarshal` methods in the provider. From what I can tell they work fine.

Additionally we're not using the unmarshalled `typedInputs` to construct the outputs, but rather use the `inputs` `resource.PropertyMap` to construct the outputs.
From that aspect, we're properly handling secrets.

Unknowns are a good point though. I'll try to double check whether this can receive unknowns if `SupportsPreview: false`.

provider/pkg/resources/cfn_custom_resource.go
flostadler: I'll double check. It's not a big deal if we stick to using the PropertyMap for Check

provider/pkg/resources/cfn_custom_resource.go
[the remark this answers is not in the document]
t0yv0: I was confused that this is called "serviceToken". This made me think of access tokens or some such. If there's precedent of calling lambda ARNs "tokens" then TIL.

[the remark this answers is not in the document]
flostadler: I also don't like the name... It's what CloudFormation called this property. I think there's precedence to stick to their naming patterns given we're trying to emulate their behavior/API

PR #1807, opened 2024-11-07
Expose CloudFormation Custom Resource Emulator Resource

flostadler:
This change exposes the new CloudFormation Custom Resource Emulator resource.
Additionally, it adds an integration test for it and makes `Check` correctly handle unknowns.

One follow up item is to translate the code example to other languages.


flostadler:

- #1805
    - #1806
        - #1807 ◀

provider/pkg/provider/provider.go
```diff
@@ -1125,7 +1135,7 @@ func (p *cfnProvider) Delete(ctx context.Context, req *pulumirpc.DeleteRequest)
 			KeepSecrets:  true,
 		})
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to parse inputs for update")
+			return nil, errors.Wrapf(err, "failed to parse inputs for delete")
```
flostadler, line 1138: Found this copy&paste error in the message. We're in the `delete` method here

provider/pkg/schema/docs/content/cfn-custom-resource.md
```diff
@@ -0,0 +1,56 @@
...
+const bucket = new aws.s3.Bucket('custom-resource-emulator');
+
+// Create a Custom Resource that invokes a Lambda function
+const cr = new aws.cloudformation.CustomResourceEmulator('cr', {
+    bucketName: bucket.id,
+    bucketKeyPrefix: 'custom-resource-emulator',
+    customResourceProperties: {
+        hello: "world"
+    },
+    serviceToken: "arn:aws:lambda:us-west-2:123456789012:function:my-custom-resource",
```
t0yv0, line 19: This lambda would be provisioned by CF, is that right?

flostadler: No, it is provisioned by the user in their Pulumi program or CloudFormation Stack.

examples/cfn-custom-resource/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    role: lambdaRole.name,
+    policyArn: policy.arn,
+});
+
+const rpa2 = new awsClassic.iam.RolePolicyAttachment("lambdaRolePolicyAttachment2", {
+    role: lambdaRole.name,
+    policyArn: awsClassic.iam.ManagedPolicies.AWSLambdaBasicExecutionRole,
+});
+
+const bucket = new awsClassic.s3.Bucket('custom-resource-emulator', {
```
t0yv0, line 35: BucketV2

examples/cfn-custom-resource/index.ts
```diff
@@ -0,0 +1,98 @@
...
+}, { dependsOn: [rpa1, rpa2] });
+
+const cfnCustomResource = new aws.cloudformation.CustomResourceEmulator('emulator', {
+    bucketName: bucket.id,
+    bucketKeyPrefix: 'custom-resource-emulator',
+    customResourceProperties: {
+        Region: amiRegion,
+        Architecture: 'HVM64',
+    },
+    serviceToken: lambdaFunction.arn,
```
t0yv0, line 65: Ahhh so serviceToken is not called lambdaFunctionARN because some custom resources are backed by something else?

t0yv0 (approved):

provider/pkg/schema/docs/content/cfn-custom-resource.md
t0yv0: This Example usage should probably have a link to how to do that.

examples/cfn-custom-resource/index.ts
flostadler: Yeah, there's also SNS backed Custom Resources: https://github.com/pulumi/pulumi-aws-native/issues/1812

They're not widely used, so we decided to skip support for them for now. It's not in scope of the CDK work

t0yv0 (approved):

examples/cfn-custom-resource/index.ts
t0yv0: The naming makes a bit more sense now.

provider/pkg/schema/docs/content/cfn-custom-resource.md
flostadler: The `Lambda Function Requirements` section explains how to do that. I don't think we should add a full Custom Resource implementation here. That would essentially be the `cfn-custom-resource` example I've added (and that's multiple files)

examples/examples_nodejs_test.go
```diff
@@ -29,6 +35,46 @@ func TestGetTs(t *testing.T) {
...
+
+	cwd := getCwd(t)
+	options := []opttest.Option{
+		opttest.LocalProviderPath("aws-native", filepath.Join(cwd, "..", "bin")),
+		opttest.YarnLink("@pulumi/aws-native"),
+	}
+	test := pulumitest.NewPulumiTest(t, filepath.Join(cwd, "cfn-custom-resource"), options...)
+	test.SetConfig(t, "amiRegion", "us-west-2")
+
+	previewResult := test.Preview(t)
```
flostadler, line 57: `Check` will receive unknowns during preview. (even if supports preview is not enabled).
I'll change `Check` to handle this, it should be relatively easy

flostadler: Done

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -241,23 +241,20 @@ func (c *cfnCustomResource) Check(ctx context.Context, urn urn.URN, randomSeed [
...
 	}
 
-	if typedInputs.StackID == nil {
+	_, hasStackID := inputs[resource.PropertyKey("stackId")]
+	if !hasStackID && typedInputs.StackID == nil {
 		// if the stack ID is not provided, we use the pulumi stack ID as the stack ID
 		inputs[resource.PropertyKey("stackId")] = resource.NewStringProperty(urn.Stack().String())
 	}
 
-	if typedInputs.CustomResourceProperties != nil {
```
flostadler, line 256: I decided to move `stringifying` the input `customResourceProperties` into the Create/Update/Delete methods.

This is an implementation detail and shouldn't really be done in `Check`. It would also complicate handling secrets and unknowns.

examples/cfn-custom-resource/ami-lookup.js
```diff
@@ -0,0 +1,103 @@
...
+    await sendResponse(event, context, responseStatus, responseData);
+};
+
+// Check if the image is a beta or rc image. The Lambda function won't return any of those images.
+function isBeta(imageName) {
+    return imageName.toLowerCase().indexOf("beta") > -1 || imageName.toLowerCase().indexOf(".rc") > -1;
+}
+
+// Send response to the pre-signed S3 URL 
+async function sendResponse(event, context, responseStatus, responseData) {
```
t0yv0, line 57: I am curious here, when people write lambdas to power custom resources in CloudFormation, do they need to take care of this functionality in user code or is there some library or framework that users can call in CF?

t0yv0 (approved):

examples/examples_nodejs_test.go
```diff
@@ -29,6 +35,46 @@ func TestGetTs(t *testing.T) {
...
+
+	test.SetConfig(t, "amiRegion", "us-east-1")
+	upResult = test.Up(t)
+	t.Logf("#%v", upResult.Summary)
+	crossTest(t, upResult.Outputs)
+
+	previewResult = test.Preview(t)
+	assertpreview.HasNoChanges(t, previewResult)
+
+	test.Destroy(t)
```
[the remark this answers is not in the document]
flostadler: Yeah it is, but I wanted to be explicit about this being one of the tested aspects. Happy to remove it though

examples/cfn-custom-resource/ami-lookup.js
flostadler: This needs to be part of the user code in the lambda function.

When using Inline Code in CloudFormation you can use the `cfn-response` module if it's a node Lambda. That's automatically included in the Lambda Zip when using inline code.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html#cfn-lambda-function-code-cfnresponsemodule-source

provider/pkg/schema/docs/content/cfn-custom-resource.md
flostadler: Yeah, I can add that once it's merged into master :)

examples/cfn-custom-resource/ami-lookup.js
t0yv0: You could use it in the example too presumably? https://www.npmjs.com/package/cfn-response seems available. No big deal though

flostadler: That's not the official version. Somebody just extracted it and published it on Lambda. I wouldn't wanna rely on this in-official version for the example

PR #1808, opened 2024-11-07
Add utility functions for supporting CloudFormation CustomResources

flostadler:
This change adds utility methods needed for supporting CloudFormation custom resources.

The main part of this change is a method for deeply converting all primitive values in a `map[string]interface{}` to strings. This is needed because CloudFormation has this behavior: https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1037

### DIFF
--- a/provider/pkg/naming/convert.go
+++ b/provider/pkg/naming/convert.go
@@ -288,13 +288,13 @@ func cfnValueToSdk(value interface{}) interface{} {
 	}
 }
 
-// ToStringifiedMap creates a copy of the input map with all primitive values converted to strings.
-func ToStringifiedMap(value map[string]interface{}) map[string]interface{} {
+// ToStringifiedMap creates a copy of the input map with all deeply nested primitive values converted to strings.
+func ToStringifiedMap[K comparable](value map[K]interface{}) map[K]interface{} {
 	if value == nil {
 		return nil
 	}
 
-	result := map[string]interface{}{}
+	result := map[K]interface{}{}
 	for k, v := range value {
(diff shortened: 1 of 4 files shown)


flostadler:

- #1805
    - #1808 ◀
        - #1806
            - #1807

provider/pkg/naming/convert.go
```diff
@@ -288,6 +288,47 @@ func cfnValueToSdk(value interface{}) interface{} {
...
+
+func primitivesToString(value interface{}) interface{} {
+	if value == nil {
+		return nil
+	}
+
+	switch reflect.TypeOf(value).Kind() {
+	case reflect.Map:
+		valueMap, ok := value.(map[string]interface{})
+		if !ok {
```
corymhall, line 312: Under what circumstance would this happen (`!ok`)? Should we throw an error here instead?

provider/pkg/naming/convert.go
```diff
@@ -288,6 +288,47 @@ func cfnValueToSdk(value interface{}) interface{} {
...
+			return value
+		}
+		result := map[string]interface{}{}
```
corymhall, line 315: Could we just call `ToStringifiedMap` instead to avoid duplicating the code?

corymhall (approved):

provider/pkg/naming/convert.go
flostadler: In this case it shouldn't happen because CloudFormation only supports strings as keys.

Throwing sounds more apt in this case. Will change it accordingly!

Issue #1186, opened 2023-11-23
Intermitent errors creating aws-native:ec2:SubnetRouteTableAssociation


mjeffryes:
### What happened?

Pulumi-cdk has been experiencing flaky test runs due to the following error:

```
 aws-native:ec2:SubnetRouteTableAssociation (VPCPublicSubnet1RouteTableAssociation0B0896DC):
      error: reading resource state: reading resource state: operation error CloudControl: GetResource, https response error StatusCode: 400, RequestID: fca75fd6-63d7-4569-8853-49f8558a8242, ResourceNotFoundException: AWS::EC2::SubnetRouteTableAssociation Handler returned status FAILED: No route tables Found with association rtbassoc-03189a84b8caa818f (HandlerErrorCode: NotFound, RequestToken: 81dddeca-f9da-44a5-9b93-9ff127dfb1c4)
```

This error is encountered here: https://github.com/pulumi/pulumi-aws-native/blob/master/provider/pkg/provider/provider.go#L803 when cloud control fails to find the resource it has just created. 

We'll need to work this down to a repro case to send to the CC API maintainers; we can also look into a workaround in aws-native (eg. adding our own retries for this particular case).

### Example

https://github.com/pulumi/pulumi-cdk/tree/main/examples/alb is our best repro case at the moment. 

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
I have been looking a this a little bit, it appears notoriously difficult to reproduce locally.  I have boiled down the relevant resource-set from the CDK examples to something minimal as follows:


```typescript

import * as aws from "@pulumi/aws-native";

const vpc = new aws.ec2.Vpc("my-vpc", {
    cidrBlock: "10.0.0.0/16",
    enableDnsHostnames: true,
    enableDnsSupport: true,
    instanceTenancy: "default",
    tags: [
        {
            key: "Name",
            value: "fargatestack/MyVpc"
        }
    ]
});


const mySubnet = new aws.ec2.Subnet("my-subnet", {
    availabilityZone: "us-west-2a",
    cidrBlock: "10.0.0.0/18",
    mapPublicIpOnLaunch: true,
    tags: [
        {
            key: "aws-cdk:subnet-name",
            value: "Public"
        },
        {
            key: "aws-cdk:subnet-type",
            value: "Public"
[...]

PR #1809, opened 2024-11-07
Retry GetResource NotFound when creating

t0yv0:
Introduces a retry for NotFound errors from GetResource executed right after Create. 

Based on the logs from our CI runs we suspect eventual consistency in AWS may cause some resources to fail with NotFound in Get even after WaitForResourceOpCompletion succeeded.

Relates: pulumi/pulumi-aws-native#1186

### DIFF
--- a/provider/pkg/client/client.go
+++ b/provider/pkg/client/client.go
@@ -84,10 +84,18 @@ func (c *clientImpl) Create(ctx context.Context, typeName string, desiredState m
 		return nil, nil, errors.New("received nil identifier while awaiting completion")
 	}
 
+	// Pick the getResource method. If there was an error from awaiting completion, simply try once. If there were
+	// no errors however, try multiple times to tolerate occasional GetResource NotFound errors that arise due to
+	// eventual consistency.
+	getResource := c.api.GetResource
+	if waitErr == nil {
+		getResource = c.getResourceRetryNotFound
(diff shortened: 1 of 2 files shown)


corymhall (approved):

provider/pkg/client/client.go
```diff
@@ -152,3 +165,39 @@ func (c *clientImpl) Delete(ctx context.Context, typeName, identifier string) er
...
 	return err
 }
+
+// Wraps c.api.GetResource with Not Found error retry logic to try to compensate for eventual consistency issues for
+// newly created resources.
+func (c *clientImpl) getResourceRetryNotFound(
+	ctx context.Context,
+	typeName, identifier string,
+) (map[string]interface{}, error) {
+	retryBackoff := retry.NewExponentialJitterBackoff(30 * time.Second)
```
flostadler, line 175: I'd reduce the max duration here a bit. We'd now wait up to ~60 seconds. That seems a bit too long to me.

Eventual consistency delays for those services are usually in the millis.

flostadler (commented):
LGTM

provider/pkg/client/client.go
t0yv0: I'll send a quick follow-up PR, this one merged already.

PR #1810, opened 2024-11-07
Adjust GetResource NotFound backoff

t0yv0:
GetResource NotFound backoff introduced in https://github.com/pulumi/pulumi-aws-native/pull/1809 allowed long time wait periods up to 30 seconds. This change makes the max wait period 3 seconds at this is more appropriate for AWS eventual consistency propagation delays this backoff tries to compensate for.


provider/pkg/client/client_test.go
```diff
@@ -365,3 +372,48 @@ func (m *mockAPI) GetResourceRequestStatus(ctx context.Context, requestToken str
...
+func TestGetResourceRetryNotFoundRetrySettings(t *testing.T) {
+	ci := &clientImpl{}
+	millisecondDelays := []float64{}
+	attempts, backoff := ci.getResourceRetryNotFoundRetrySettings()
+	for attempt := 0; attempt < attempts; attempt++ {
+		delay, err := backoff.BackoffDelay(attempt, nil)
+		require.NoError(t, err)
+		millisecondDelays = append(millisecondDelays, float64(delay/time.Millisecond))
+	}
+	autogold.Expect([]float64{853, 437, 3000}).Equal(t, millisecondDelays)
```
t0yv0, line 385: These are millisecond values for the attempt 0, 1, 2 suggested by backoff. Jitter makes them random in prod, but in testing the randomness is eliminated (see below).

corymhall (approved):

PR #1823, opened 2024-11-13
Additional backof delay for retrying Resource.Get Not Found

t0yv0:
It appears that the CDK test suite is still failing, so this change adjusts the backoff/retry on Resource.Get to be more generous and allow more time. 

https://github.com/pulumi/pulumi-cdk/actions/runs/11818972903/job/32927805461?pr=212


blampe (approved):
Worth a try.

Issue #1816, opened 2024-11-11
Handle missing auto-naming constraints


flostadler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The CloudFormation schema provided by AWS, which we use to generate the AWS Native provider metadata, lacks resource name length constraints for many AWS services. This is particularly problematic for resources like `lambda.Function` and `iam.Role` where AWS enforces strict length limitations (e.g., 64 characters for Lambda Role names) at the API level.

When users rely on the provider's auto-naming feature, deployments can fail unexpectedly because the provider isn't aware of these AWS-imposed limits. Since these constraints aren't present in the CloudFormation schema, the provider cannot validate or handle resource names appropriately before deployment.

### Proposal
Enhance provider metadata with length constraints for common AWS resources:
   - Manually add known length limits for popular resources (e.g., IAM Roles, Lambda Functions)
   - Implement intelligent name handling for resources exceeding limits
     - When auto-naming is enabled, trim long names while preserving uniqueness
       - Example: For a 100-character resource name with a 64-character limit, trim to 56 characters and append the random suffix (similar to AWS CloudFormation's approach)

Considerations:
   - This change should be non-breaking as resources exceeding AWS limits are currently undeployable with auto-naming. But it cannot rely only on the CloudFormation schema for constraint validation as it might be incorrect
       - Existing resources should retain their current names from state
   - Should document the auto-trimming behavior for transparency

### Affected area/feature

- aws-native
- pulumi-cdk

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/2191

PR #1828, opened 2024-11-14
Add autonaming overlay

corymhall:
Most resources have some limits on what the resource name can be.
Unfortunately a lot of those limits are not currently stored in the
CloudFormation schema.

This PR introduces a new schema overlay where we can manually store
min/max length constraints for resource names.

This is the first step in addressing #1816. I will follow this up with
another PR to trim names to fit within the constraints.

re #1816, re https://github.com/pulumi/pulumi-cdk/issues/62


aws-cfn-autoname-overlay.json
```diff
@@ -0,0 +1,34 @@
...
+    "AWS::IAM::Role": {
+      "RoleName": {
+        "minLength": 1,
+        "maxLength": 64
+      }
+    }, 
+    "AWS::Lambda::Function": {
+      "FunctionName": {
+        "minLength": 1,
+        "maxLength": 140
```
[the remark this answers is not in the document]
corymhall: Wow good catch! We want 64 here since this will only be used in creating an auto name.

examples/examples_nodejs_test.go
```diff
@@ -141,6 +142,21 @@ func TestNamingConventions(t *testing.T) {
...
 
+func TestAutoNamingOverlay(t *testing.T) {
+	var buf bytes.Buffer
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir:           filepath.Join(getCwd(t), "autonaming-overlay"),
+			Stderr:        &buf,
+			ExpectFailure: true,
+			ExtraRuntimeValidation: func(t *testing.T, stackInfo integration.RuntimeValidationStackInfo) {
+				assert.Contains(t, buf.String(), "is too large to fix max length constraint of 64")
```
flostadler, line 153: Out of scope, but shouldn't the error message here be "... too large to **fit** max length constraint". Maybe it's a non-native speaker issue on my end.

I can cut a PR if we think this is wrong

provider/cmd/pulumi-gen-aws-native/main_test.go
```diff
@@ -53,6 +54,85 @@ func TestCleanDir(t *testing.T) {
...
+
+		mergeAutoNaming(schema, overlay)
+
+		assert.Equal(t, &jsschema.Schema{
+			MaxLength: jsschema.Integer{Val: 64},
+			MinLength: jsschema.Integer{Val: 1},
+		}, schema.Properties["RoleName"])
+	})
+
+	t.Run("overlay does not overwrites", func(t *testing.T) {
```
flostadler, line 85:
```suggestion
	t.Run("overlay does not overwrite", func(t *testing.T) {
```

flostadler (approved):
Nice!

examples/examples_nodejs_test.go
corymhall: yeah I think it was just a typo

PR #1831, opened 2024-11-14
generate sdks

corymhall:


provider/pkg/autonaming/application.go
```diff
@@ -70,6 +91,19 @@ func getDefaultName(
 	if autoNamingSpec.MaxLength > 0 {
 		left := autoNamingSpec.MaxLength - len(prefix) - namingTrivia.Length()
 
+		if left <= 0 && autoTrim {
+			autoTrimMaxLength := autoNamingSpec.MaxLength - namingTrivia.Length() - randomSuffixMinLength
+			if autoTrimMaxLength <= 0 {
+				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
```
t0yv0, line 97: So surprised to see Fortran-style 1-based indices here but I think this checks out. This is right.

provider/pkg/autonaming/application_test.go
```diff
@@ -73,7 +74,123 @@ func Test_getDefaultName(t *testing.T) {
...
+			comparison: equals(resource.NewStringProperty("oldName")),
+		},
+		{
+			resourceName: "myReallyLongName",
+			name:         "Autoname with constraints on max length",
+			maxLength:    12,
+			autoNameConfig: AutoNamingConfig{
+				AutoTrim: true,
+			},
+			comparison: within(12, 12),
```
t0yv0, line 138: Any chance we can see `equals(resource.NewStringProperty(".."))` here or it's omitted because it is random?

t0yv0:
Added https://github.com/pulumi/pulumi-aws-native/issues/1833 but out of scope of this PR.... Looks like we are consuming GetVariables still in this provider.

t0yv0 (approved):
LGTM 🚢

t0yv0:
Assuming here https://github.com/pulumi/pulumi/discussions/17592 is not quite ready for prime time so we need this now.

provider/pkg/autonaming/application.go
```diff
@@ -101,3 +135,25 @@ func getDefaultName(
 
 	return resource.NewStringProperty(random), nil
 }
+
+// trimName will trim the prefix to fit within the max length constraint.
+// It will cut out part of the middle, keeping the beginning and the end of the string.
```
flostadler, line 140: I think this is the right thing to do given the current constraints. But it could also be rather surprising to users.

For example, I often split my resource names with hyphens. This could generate some rather odd names.

But given that https://github.com/pulumi/pulumi/discussions/17592 is not ready yet, I think this is a good choice.
If we see the need for it, we could also expose more configuration options for autonaming in the future.

flostadler (approved):
LGTM

provider/pkg/provider/provider.go
```diff
@@ -519,6 +520,14 @@ func (p *cfnProvider) Configure(ctx context.Context, req *pulumirpc.ConfigureReq
 		metadata.CfnCustomResourceToken: resources.NewCfnCustomResource(p.name, s3Client, lambdaClient),
 	}
 
+	if autoNaming, ok := vars["aws-native:config:autoNaming"]; ok {
+		var autoNamingConfig autonaming.AutoNamingConfig
```
flostadler, line 524: Can we test the unmarshalling? It seems to be untested right now

provider/pkg/autonaming/application_test.go
corymhall: I think I might be able to do a regex check because only the end should be random.

corymhall: Updated

provider/pkg/provider/provider.go
corymhall: Updated!

PR #1853, opened 2024-11-25
chore: update internal pulumi dependencies

t0yv0:
This change is based on a Renovate-prepared PR (https://github.com/pulumi/pulumi-aws-native/pull/1852) and a small manual fix added on top.


sdk/python/pulumi_aws_native/databrew/recipe.py
```diff
@@ -118,9 +118,9 @@ def __init__(__self__,
                 "action": {
                     "operation": "EXTRACT_PATTERN",
                     "parameters": {
-                        "sourceColumn": "Consulate",
+                        "source_column": "Consulate",
```
t0yv0, line 121: This is a little suspect. I wonder if the change is right or it is a regression.

sdk/python/pulumi_aws_native/iam/role.py
```diff
@@ -469,7 +469,7 @@ def __init__(__self__,
             actions=[{
                 "job_name": my_job.id,
                 "arguments": {
-                    "--job-bookmark-option": "job-bookmark-enable",
+                    "__job_bookmark_option": "job-bookmark-enable",
```
t0yv0, line 472: This does not look right. Looks like a regression. Will need to confirm.

provider/go.mod
```diff
@@ -38,6 +38,11 @@ require (
 	google.golang.org/protobuf v1.35.1
 )
 
+require (
+	github.com/pulumi/pulumi/pkg/v3 v3.134.0
+	github.com/pulumi/pulumi/sdk/v3 v3.134.0
+)
+
```
blampe, line 45: Your local ref was pointing at https://github.com/pulumi/pulumi-aws-native/pull/1844 instead of https://github.com/pulumi/pulumi-aws-native/pull/1852. I'll rebase you and look into making the branch name more random to prevent this.

blampe (approved):

.github/workflows/build.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         lfs: true
     - id: version
       name: Set Provider Version
-      uses: pulumi/provider-version-action@v1
+      uses: pulumi/provider-version-action@7c54f136703646f7d6eaa3d3b3c877e5a805d6ab # v1
```
t0yv0, line 55: Is this going to churn with ci-mgmt updates?

PR #1891, opened 2024-12-06
Enable preview for create and update

### LINKED ISSUE #1141: KMS keyPolicy update on preview wrongly indicates that KMS key arn will change
### What happened?

If you have a multi-zone key (primary key + replica key) and you update the `keyPolicy` attribute of the primary key, the `pulumi preview` wrongly indicates that the `primaryKeyArn` of the replica key will be updated.

The same applies for the KMS key Alias resource, that indicates that the `targetKeyId` will be updated (indicating that the alias moves to another KMS key).

As KMS keys are quite important, you might scare as developer and stop to proceed if Pulumi indicates such wrong information.

### Example

With the following program:

```
import * as awsClassic from "@pulumi/aws";

const { partition } = await awsClassic.getPartition();
const { accountId } = await aws.getAccountId();
const { name: regionName } = await awsClassic.getRegion();

const awsEuWest1Provider = new aws.Provider("euWest1", { ...aws.config, region: "eu-west-1"});
const defaultKeyPolicy = awsClassic.iam.getPolicyDocumentOutput({
    statements: [
        {
            sid: "EnableAllPermissionsForRootAccount",
            actions: ["kms:*"],
            resources: ["*"],
            effect: "Allow",
            principals: [{
                type: "AWS",
                identifiers: [`arn:${partition}:iam::${accountId}:root`],
            }]
        }]
});

const queueKey = new aws.kms.Key("kms-encrypt-queues", {
    name: "testing",
    enabled: true,
    keySpec: "SYMMETRIC_DEFAULT",
    multiRegion: true,
    pendingWindowInDays: 7,
    keyPolicy: defaultKeyPolicy.then(o => o.json),
});

const queueKeyEuWest1Replica = new aws.kms.ReplicaKey("kms-encrypt-queues-replica-eu-west-1", {
    primaryKeyArn: queueKey.arn,
    keyPolicy: defaultKeyPolicy.then(o => o.json),
    enabled: true,
    pendingWindowInDays: 7,
}, { parent: queueKey, provider: awsEuWest1Provider });

const queueKeyAlias = new aws.kms.Alias("kms-encrypt-queues-alias", {
    aliasName: `alias/testing/1`,
    targetKeyId: queueKey.arn
}, { parent: queueKey });

```

1. Run the code above with pulumi up in any region besides `eu-west-1`.
2. Adjust the policy within the variable `defaultKeyPolicy` to any new value.
3. Run `pulumi preview`, get the output below (which is wrong, because the arn of the key remains intact).
4. If you ran `pulumi up -y --skip-preview`, you see that pulumi does not update the `primaryKeyArn` or `targetKeyId`, as updating a policy changes nothing in that regard.

```
     pulumi:pulumi:Stack              your-stack-name              
 ~   ├─ aws-native:kms:Key            kms-encrypt-queues                         update      [diff: ~keyPolicy]
 ~   │  ├─ aws-native:kms:Alias       kms-encrypt-queues-alias                   update      [diff: ~targetKeyId]
 +-  │  └─ aws-native:kms:ReplicaKey  kms-encrypt-queues-replica-eu-west-1       replace     [diff: ~keyPolicy,primaryKeyArn]
```

Solution: only indicate changes in the primary key arn if the key will be/is replaced.

### Output of `pulumi about`

```
NAME                VERSION
@pulumi/aws         6.5.0
@pulumi/aws-native  0.81.0
@pulumi/pulumi      3.89.0
```

### Additional context

_No response_

### Contributing

corymhall:
Currently if you update a resource, all outputs are shown as computed
during preview. This means that downstream resources might show a
replacement in the plan if they are using one of those output values.

There are some outputs that we know will not change unless the resource
is replaced (I'm referring to these as \"stable\" outputs). For stable
outputs we can copy the value from the state in order to show an
accurate preview.

The problem is that the CCAPI schema has no way of determining
programmatically which outputs are stable and which are not. Because of
this, this PR introduces a heuristic to determine if an output is a
stable output.

- Is the value a `readOnlyProperty` in the schema? This indicates that
  it is a computed output property
- Is the property the resource `id`, `arn` or `name`? If the id, arn, or
  name is a computed property then we can assume that it is a stable
  property. I don't know of any cases where these change outside of a
  resource replacement.

This is not 100% coverage of stable properties, but it should get us
most of the impactful properties. To get close to 100% coverage we will
probably need a schema overlay where we can contribute to mapping these
stable properties for each resource.

Note: as part of this I had to add `readOnly` properties to `metadata.json`

closes #1141, re #1892


provider/pkg/provider/provider.go
```diff
@@ -1077,6 +1102,30 @@ func (p *cfnProvider) Update(ctx context.Context, req *pulumirpc.UpdateRequest)
...
+		}
+
+		previewState, err := plugin.MarshalProperties(
+			outputs,
+			plugin.MarshalOptions{Label: fmt.Sprintf("%s.checkpoint", label), KeepSecrets: true, KeepUnknowns: true, SkipNulls: true},
+		)
+		if err != nil {
+			return nil, err
+		}
+		return &pulumirpc.UpdateResponse{Properties: previewState}, nil
```
t0yv0, line 1126: Id is omitted because it's unknown?

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+	case inputValue.IsArray() && (prop.Type == "array" || prop.Items != nil):
+		var items []resource.PropertyValue
+		for _, item := range inputValue.ArrayValue() {
+			items = append(items, previewOutputValue(item, types, prop.Items, arrayReadOnly(readOnly)))
+		}
+		return resource.NewArrayProperty(items)
+	case inputValue.IsObject() && strings.HasPrefix(prop.Ref, "#/types/"):
+		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
+		if t, ok := types[typeName]; ok {
+			v := previewResourceOutputs(inputValue.ObjectValue(), types, t.Properties, readOnly)
```
t0yv0, line 204: This call here is no longer addressing a resource, but a nested object type.

provider/pkg/provider/provider.go
```diff
@@ -1077,6 +1102,30 @@ func (p *cfnProvider) Update(ctx context.Context, req *pulumirpc.UpdateRequest)
...
+		if _, ok := p.customResources[resourceToken]; ok {
+			outputs = PreviewOutputs(newInputs, nil, nil, nil)
+		} else {
+			spec, hasSpec := p.resourceMap.Resources[resourceToken]
+			if !hasSpec {
+				return nil, errors.Errorf("Resource type %s not found", resourceToken)
+			}
+			resourceTypeName := urn.Type().Name()
+			outputs = PreviewOutputs(newInputs, p.resourceMap.Types, spec.Outputs, spec.ReadOnly)
+			outputs = populateStableOutputs(newInputs, oldState, spec.ReadOnly, resourceTypeName)
```
t0yv0, line 1116: Why is populateStableOutputs not a part of PreviewOutputs implementation?

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+	types map[string]metadata.CloudAPIType,
+	outputs map[string]schema.PropertySpec,
+	readOnly []string,
+) resource.PropertyMap {
+	result := resource.PropertyMap{}
+	for name, prop := range outputs {
+		key := resource.PropertyKey(name)
+		if inputValue, ok := inputs[key]; ok {
+			result[key] = previewOutputValue(inputValue, types, &prop.TypeSpec, filterReadOnly(readOnly, name))
+			// if the property is a readOnly property, then it's an output only value
```
t0yv0, line 103: Could you explain this bit? Why are readOnly vs non-read-only values different handled differently when trying to predict outputs? One gets omitted but the other gets an Unknown value? Are those perhaps treated de-facto in the same way by the engine?

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+func isReadOnly(readOnly []string, key string) bool {
+	for _, prop := range readOnly {
+		if naming.ToSdkName(prop) == key {
+			return true
+		}
+	}
+	return false
+}
+
+func filterReadOnly(readOnly []string, key string) []string {
```
t0yv0, line 162: This does a little bit more than filtering it seems.

t0yv0:
Let's discuss briefly tomorrow, are recursive cases strictly necessary - perhaps worth shipping a smaller much simpler change first to test the waters (covering ARN properties) and then extend as needed? Thanks.

provider/pkg/provider/previewOutputs.go
corymhall: During preview we have the inputs from the user and need to project them into outputs. We loop over all the output types and if we find an input that matches we carry over the value. If there is an output type where the value is not found in the inputs then it could be one of two cases

1. The output is a readonly type so it is not expected to be found in the inputs and is a computed value.
2. The output is not a readonly type which means that it is an input, but the user has just not specified it.

For these cases I assumed that we should just leave it as omitted, but should we just mark these as computed too? I suppose some of them could be fields that have default values which means that the value would be computed. The CCAPI schema isn't verbose enough to tell us whether an input is computed or not so should we just assume it is?

provider/pkg/provider/provider.go
corymhall: Mainly because we want to use `PreviewOutputs` on both `Create` & `Update`, but we only use `populateStableOutputs` on `Update`. If you think it would be better to combine it all in `PreviewOutputs` and just decide what to do  based on whether `oldState` is provided I could do that as well, I don't have a strong preference either way.

provider/pkg/provider/previewOutputs.go
corymhall: I can rename it to something else, but why do you think it's hard to be sure this always does the right thing?

provider/pkg/provider/provider.go
corymhall: Id isn't part of the UpdateResponse

provider/pkg/provider/previewOutputs.go
corymhall: I'll add a comment

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+// It takes the provided resource inputs and maps them to the outputs shape, adding unknowns
+// for all properties that are not defined in inputs
+func PreviewOutputs(
+	inputs resource.PropertyMap,
+	types map[string]metadata.CloudAPIType,
+	props map[string]schema.PropertySpec,
+	readOnly []string,
+) resource.PropertyMap {
+	result := resource.PropertyMap{}
+	// Then this is an Extension resource which has all outputs in an "outputs" property
```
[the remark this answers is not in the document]
corymhall: Yeah maybe I should just split it into separate functions.

provider/pkg/provider/previewOutputs.go
t0yv0: Ah that's a helpful. I think your code does the right thing. I suspect omitting something is treated similar to having it as Null, but sending an Unknown is different from Null as it signals something might change during update. 

1. The output is a readonly type so it is not expected to be found in the inputs and is a computed value.
This case should be unknown. If we leave it undefined (null), we're being over-confident in that the provider will not specify it.

2. The output is not a readonly type which means that it is an input, but the user has just not specified it.
This case should be left undefined. We are being confident here that the provider will not modify this.

provider/pkg/provider/provider.go
t0yv0: Neither do I, feel free to ignore, it is a matter of style, this organization also makes sense. Just looked a little weird to me (also, different capitalization, is PreviewOutputs exported but populateStableOutputs is not)?

provider/pkg/provider/previewOutputs_test.go
```diff
@@ -0,0 +1,416 @@
...
+import (
+	"testing"
+
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/metadata"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrevierOutputs(t *testing.T) {
```
t0yv0, line 12: Typo in test name? TestPreviewOutputs?

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
+		if t, ok := types[typeName]; ok {
+			v := previewResourceOutputs(inputValue.ObjectValue(), types, t.Properties, readOnly)
+			return resource.NewObjectProperty(v)
+		}
+	case inputValue.IsObject() && prop.AdditionalProperties != nil:
+		inputObject := inputValue.ObjectValue()
+		result := resource.PropertyMap{}
+		for name, value := range inputObject {
+			p := value
```
t0yv0, line 211: If I insert a `panic("!")` here and run tests, this line does not get hit. This could be interesting to test out since I think recurring into the map element would need some kind of transformation on the `readOnly` parameter maybe? or if it's not needed it's good to verify.

t0yv0:
This made me curious to poke at codecov. Besides the map case which is certainly interesting, there is this bit that's potentially interesting to cover:

```
+ 	// we can't handle arrays because we don't know which item in the array
+ 	// the value corresponds to
+ 	if isArrayProperty(propName) {
+ 		return false
+ 	}
+ 
```

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,218 @@
...
+			result[name] = previewOutputValue(p, types, prop.AdditionalProperties, readOnly)
+		}
+		return resource.NewObjectProperty(result)
+	}
+	// fallback to assuming input is the same as the output
+	return inputValue
```
t0yv0, line 217: This transformation is incomplete in handling the possible grammar cases of resource.PropertyValue, the ones I'm thinking about may be worth calling out:

- NewSecretProperty
- NewOutputProperty

More of these are in https://github.com/pulumi/pulumi/blob/5d73d64185ad32da8787a0e06875ecc59a912b11/sdk/go/common/resource/properties.go#L246

Also noting that the fallback is a little aggressive as it assumes the input value will be left unmodified. 

Perhaps there's something in the provider that assures us this code will not encounter NewSecretProperty or NewOutputProperty, but if it does, values wrapped in NewSecretProperty will not receive unknown marks for readOnly sub-properties but will be overly confident.

provider/pkg/provider/previewOutputs_test.go
```diff
@@ -0,0 +1,416 @@
...
+				"name": "my-access-point",
+				"objectLambdaConfiguration": map[string]interface{}{
+					"allowedFeatures":          []string{"GetObject-Range"},
+					"cloudWatchMetricsEnabled": true,
+				},
+			}),
+			map[string]metadata.CloudAPIType{
+				"aws-native:s3objectlambda:AccessPointAlias": {
+					Properties: map[string]schema.PropertySpec{
+						"Status": {
```
t0yv0, line 25: This one is actually "status" in Pulumi schema, is there a reason we are capitalizing it here?

provider/pkg/provider/previewOutputs.go
t0yv0: That's a good point, I had to look at it with fresh eyes this morning, I think it's good! It does the right thing. readOnly properties are specified using paths into conforming objects, and jumping into `#ref` properties does the right thing. You have it tested too. I noticed some wrinkle inconsistencies that I added some more comments about but I'm comfortable with this now, thank you!

t0yv0:
Ah "Note: as part of this I had to add readOnly properties to metadata.json" - that's a separate piece. Let's maybe get that in as a separate quick PR and figure out the capitalization there? Thanks a ton!

provider/pkg/provider/previewOutputs.go
corymhall: We do nothing with default values. If there is an input/output property call `foo` and the user does not populate it then the input will be empty. The output though is populated as a result of reading from CCAPI so if AWS returns a default value the output will have that default value.

corymhall:

- #1898
    - #1891 ◀

provider/pkg/provider/previewOutputs.go
corymhall: I can add handling for secrets and outputs by just unwrapping them and calling `previewOutputValue` again with the unwrapped value.

As a side note, I noticed that in `aws-native` when we use `plugin.UnmarshalProperties` we pass `KeepUnknowns: true, KeepSecrets: true`. It doesn't seem like this is consistent across providers, do you know why?

provider/pkg/provider/previewOutputs.go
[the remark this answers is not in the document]
corymhall: I moved these to be methods on the `CustomResource` type since that seems to be the only good way of telling them apart.

provider/pkg/provider/previewOutputs.go
corymhall: I opted to leave it as is. If we mark every output property as computed it could make the preview very verbose since properties that might not be populated will show as computed.

provider/pkg/provider/previewOutputs.go
corymhall: Added some tests and also added some additional logic related to processing `additionalProperties`.

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,211 @@
...
+							outputFromInput.ObjectValue(),
+							outputFromState.ObjectValue(),
+							[]string{strings.Join(props[1:], "/")},
+							resourceTypeName,
+						))
+					}
+				}
+			} else {
+				key := resource.PropertyKey(readOnlyProp)
+				if output, ok := outputsFromPriorState[key]; ok {
```
t0yv0, line 88: This branch will not consider outpusFromInputs? Why?

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,211 @@
...
+// (i.e. they will only change if the resource is replaced). Because of this we
+// cannot blanket copy all outputs from the prior state. The ones that are not stable
+// should remain computed since they could be changing.
+//
+// Instead, we use some heuristics to determine which outputs are stable and which are not.
+// We are making the assumption that Name, Id, and Arn are stable outputs. Sometimes resources
+// will have `name`, `id`, and `arn` properties, and sometimes they will have `resourceName`, `resourceId`,
+// and `resourceArn` properties.
+func populateStableOutputs(
+	outputsFromInputs resource.PropertyMap,
```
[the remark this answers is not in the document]
t0yv0: Like relying on modifying the first parameter is highly unexpected.

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		if isStableOutput(readOnlyProp, resourceTypeName) {
+			// nested property. Sometimes the resource Arn is in a nested property
+			if strings.Contains(readOnlyProp, "/") {
+				props := strings.Split(readOnlyProp, "/")
+				current := props[0]
+				key := resource.PropertyKey(current)
+				if outputFromInput, ok := outputsFromInputs[key]; ok {
+					if outputFromState, ok := outputsFromPriorState[key]; ok {
+						outputsFromInputs[key] = resource.NewObjectProperty(populateStableOutputs(
```
t0yv0, line 78: Will this only fire up when both inputs and prior state is present? Curious if this will exclude the case when an input is present but output not.. Ah ok that may be expected because this is only working with 'stable' info propagation. 

How about when a state entry is present but the input entry is not? Why will it skip this?

provider/pkg/provider/provider.go
```diff
@@ -1077,6 +1102,30 @@ func (p *cfnProvider) Update(ctx context.Context, req *pulumirpc.UpdateRequest)
...
+		if customResource, ok := p.customResources[resourceToken]; ok {
+			outputs = customResource.PreviewCustomResourceOutputs()
+		} else {
+			spec, hasSpec := p.resourceMap.Resources[resourceToken]
+			if !hasSpec {
+				return nil, errors.Errorf("Resource type %s not found", resourceToken)
+			}
+			resourceTypeName := urn.Type().Name()
+			outputs = previewResourceOutputs(newInputs, p.resourceMap.Types, spec.Outputs, spec.ReadOnly)
+			outputs = populateStableOutputs(newInputs, oldState, spec.ReadOnly, resourceTypeName)
```
t0yv0, line 1116: I still don't understand this code because when read naively it seems that populateStableOutputs discards the result of previewResourceOutputs and overwrites outputs. 

Should it be instead something like this?

```populateStableOutputs(newInputs, oldState, spec.ReadOnly, resourceTypeName, &outputs)```

I expect populateStableOutputs complement the work that previewResourceOutputs has done to keep populating more outputs (rewriting)?

File: provider/pkg/provider/previewOutputs.go [hunk 15]
```diff
@@ -0,0 +1,211 @@
...
+		return resource.NewArrayProperty(items)
+	case inputValue.IsObject() && strings.HasPrefix(prop.Ref, "#/types/"):
+		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
+		if t, ok := types[typeName]; ok {
+			v := previewResourceOutputs(inputValue.ObjectValue(), types, t.Properties, readOnly)
+			return resource.NewObjectProperty(v)
+		}
+	case inputValue.IsObject() && prop.AdditionalProperties != nil && (prop.AdditionalProperties.Ref != "" || prop.AdditionalProperties.Items != nil):
+		return previewOutputValue(inputValue, types, prop.AdditionalProperties, readOnly)
+	case inputValue.IsObject() && prop.AdditionalProperties != nil && prop.AdditionalProperties.Ref == "":
```
t0yv0, line 200: I think this case is good but needs to drop prop.AdditionalProperties.Ref == "" condition and handle all maps.

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,211 @@
...
+		inputObject := inputValue.ObjectValue()
+		result := resource.PropertyMap{}
+		for name, value := range inputObject {
+			p := value
+			result[name] = previewOutputValue(p, types, prop.AdditionalProperties, readOnly)
```
t0yv0, line 205: I suspect `readOnly` here is a bug, it has to be something like `mapElementReadOnly(readOnly)`, no? How does CF schema specify readOnly properties that are nested under maps? Is this a degenerate case that just does not occur?

provider/pkg/provider/previewOutputs.go
corymhall: I think I've changed my mind about a previous comment you made about this being part of `previewResourceOutputs`. This function is meant to only be called with the output of `previewResourceOutputs` so `outputsFromInputs` contains all the possible output properties already. 

If an entry is not present in the input, but is in the prior state then it will be marked as computed (as part of the `previewResourceOutputs` function.

provider/pkg/provider/previewOutputs.go
corymhall: `outputsFromInputs` will have all possible outputs properties because of `previewResourceOutputs`, so if it's a stable property and it exists in `outputsFromPriorState` then copy it over.

provider/pkg/provider/provider.go
corymhall: yes this was a bug, populateStableOutputs was supposed to take the result of previewResourceOutputs. I went ahead and refactored this so that it all happens in a single function.

provider/pkg/provider/previewOutputs.go
```diff
@@ -0,0 +1,211 @@
...
+			items = append(items, previewOutputValue(item, types, prop.Items, readOnly))
+		}
+		return resource.NewArrayProperty(items)
+	case inputValue.IsObject() && strings.HasPrefix(prop.Ref, "#/types/"):
+		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
+		if t, ok := types[typeName]; ok {
+			v := previewResourceOutputs(inputValue.ObjectValue(), types, t.Properties, readOnly)
+			return resource.NewObjectProperty(v)
+		}
+	case inputValue.IsObject() && prop.AdditionalProperties != nil && (prop.AdditionalProperties.Ref != "" || prop.AdditionalProperties.Items != nil):
```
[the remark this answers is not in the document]
corymhall: Based on our convo on additionalProperties I reverted this to what it was before. The map types don't need any special handling and won't have nested readOnly properties.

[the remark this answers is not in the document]
t0yv0: Are you sure about this? Cloud Control never has a map of objects types?  That said even if it's there it's probably super rare so we can handle later on if needed.

provider/pkg/outputs/previewOutputs.go
```diff
@@ -0,0 +1,234 @@
...
+	types map[string]metadata.CloudAPIType,
+	outputTypes map[string]schema.PropertySpec,
+	readOnlyProperties []string,
+	resourceTypeName tokens.TypeName,
+	// will only be provided during an update operation
+	outputsFromPriorState *resource.PropertyMap,
+) resource.PropertyMap {
+	outputsFromInputs := previewResourceOutputs(newInputs, types, outputTypes, readOnlyProperties)
+	if outputsFromPriorState != nil {
+		return populateStableOutputs(outputsFromInputs, *outputsFromPriorState, readOnlyProperties, resourceTypeName)
```
t0yv0, line 32: 👍

provider/pkg/outputs/previewOutputs.go
```diff
@@ -0,0 +1,234 @@
...
+// (i.e. they will only change if the resource is replaced). Because of this we
+// cannot blanket copy all outputs from the prior state. The ones that are not stable
+// should remain computed since they could be changing.
+//
+// Instead, we use some heuristics to determine which outputs are stable and which are not.
+// We are making the assumption that Name, Id, and Arn are stable outputs. Sometimes resources
+// will have `name`, `id`, and `arn` properties, and sometimes they will have `resourceName`, `resourceId`,
+// and `resourceArn` properties.
+func populateStableOutputs(
+	// outputsFromInputs has to be the output of previewResourceOutputs
```
t0yv0, line 79: 👍

provider/pkg/outputs/previewOutputs.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		return resource.NewArrayProperty(items)
+	case inputValue.IsObject() && strings.HasPrefix(prop.Ref, "#/types/"):
+		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
+		if t, ok := types[typeName]; ok {
+			v := previewResourceOutputs(inputValue.ObjectValue(), types, t.Properties, readOnly)
+			return resource.NewObjectProperty(v)
+		}
+		// AdditionalProperties (map types)
+	case inputValue.IsObject() && prop.AdditionalProperties != nil:
```
t0yv0, line 223: Nice.

t0yv0 (approved):
🚢

flostadler (approved):

examples/examples_nodejs_test.go
```diff
@@ -97,6 +97,28 @@ func TestRefreshChanges(t *testing.T) {
...
+func TestPreviewStableProperties(t *testing.T) {
+	cwd := getCwd(t)
+	options := []opttest.Option{
+		opttest.LocalProviderPath("aws-native", filepath.Join(cwd, "..", "bin")),
+		opttest.YarnLink("@pulumi/aws-native"),
+	}
+	test := pulumitest.NewPulumiTest(t, filepath.Join(cwd, "stable-outputs-preview"), options...)
+	test.SetConfig(t, "lambdaDescription", "Lambda 1")
+	defer func() {
+		test.Destroy(t)
```
flostadler, line 109: Doesn't pulumitest auto destroy by default?

examples/stable-outputs-preview/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        },
+        Action: "sts:AssumeRole",
+      },
+    ],
+  },
+});
+
+
+const bucket = new ccapi.s3.Bucket('bucket');
+const object = new aws.s3.BucketObjectv2(
```
flostadler, line 26: Meta: Kind of frustrating that many rather basic use cases require layering aws classic on top of aws native in order to make it usable 
:(

examples/stable-outputs-preview/zip.ts
```diff
@@ -0,0 +1,48 @@
+import { createWriteStream, promises as fs } from 'fs';
+import * as path from 'path';
+import * as glob from 'glob';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const archiver = require('archiver');
+
+export function zipDirectory(directory: string, outputFile: string): Promise<string> {
```
flostadler, line 8: Couldn't we do that by using an `AssetArchive`?

provider/pkg/outputs/previewOutputs.go
```diff
@@ -0,0 +1,234 @@
...
+
+// isStableOutput determines if a property is a stable output or not.
+// This uses a very conservative heuristic and does not cover all cases.
+//
+// e.g. sometimes the arn property does not contain the full resource type name
+// - `aws-native:sagemaker:ModelExplainabilityJobDefinition` has a property `jobDefinitionArn`
+// - `aws-native:securityhub:PolicyAssociation` has a property `associationIdentifier`
+//   - TODO[pulumi/aws-native#1892]: in some cases this property is the `primaryIdentifier`. Could we use that as another heuristic?
+//     a readonly property that is also a primary identifier? It doesn't catch all cases, but would catch more
+func isStableOutput(propName string, resourceTypeName tokens.TypeName) bool {
```
flostadler, line 129: Nice!

provider/pkg/resources/cfn_custom_resource.go
```diff
@@ -685,3 +685,11 @@ func sanitizeCustomResourceResponse(event *cfn.Event, response *cfn.Response) *c
 
 	return response
 }
+
+// cfn custom resources have outputs returned in a "data" property
+// since it can be any arbitrary data, we mark the entire thing as computed
+func (c *cfnCustomResource) PreviewCustomResourceOutputs() resource.PropertyMap {
```
flostadler, line 691: Small nit: this should be covered by a unit ideally

provider/pkg/resources/custom.go
```diff
@@ -21,4 +21,6 @@ type CustomResource interface {
 	Update(ctx context.Context, urn resource.URN, id string, inputs, oldInputs, state resource.PropertyMap, timeout time.Duration) (resource.PropertyMap, error)
 	// Delete removes the resource from the cloud provider.
 	Delete(ctx context.Context, urn resource.URN, id string, inputs, state resource.PropertyMap, timeout time.Duration) error
+	// PreviewCustomResourceOutputs returns the outputs of the resource based on the inputs and the output properties in the resource schema.
+	PreviewCustomResourceOutputs() resource.PropertyMap
```
flostadler, line 25: Nice!

flostadler (approved):
Love it!

PR #1893, opened 2024-12-06
Fix refresh does not show changes

### LINKED ISSUE #1796: Refresh does not show changes
### What happened?

The `Read` implementation of aws-native is incorrectly returning the old inputs instead of the newly refreshed ones.
https://github.com/pulumi/pulumi-aws-native/blob/0166a6357100a309d815a53cc4ac986c3fa96aab/provider/pkg/provider/provider.go#L1018

This has the effect that `pulumi refresh` will not show any resources to be changed. Applying this empty refresh will still modify the state though and align it with the actual values of the physical resource.
Re-running `pulumi refresh` again will now show a diff, but applying it will not change the properties, it will only align the Inputs saved in state.

### Example

```ts

const ssmParam = new aws.ssm.Parameter("my-param", {
    type: aws.ssm.ParameterType.String,
    value: "old-value",
})

export const paramName = ssmParam.name;
export const paramValue = ssmParam.value;
```

- `pulumi up` and note `paramName` output
- `pulumi stack export --file old.json`
- `aws ssm put-parameter --name $PARAM_NAME --value "new-value" --type "String" --overwrite`
- `pulumi refresh`: Observe no changes being shown
- `pulumi stack export --file new.json`
- `diff old.json new.json`: Observe that value of the SSM parameter has been updated, but the inputs are still the old ones
- `pulumi refresh`: Shows that the value is changing from "new-value" to "old-value"
- `pulumi stack export --file new2.json`
- `diff new.json new2.json`: The inputs have been aligned now

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

corymhall:
This PR fixes an issue where the `Read` method was returning the
previous (oldState) inputs which meant that refresh operations never
showed any changes.

fixes #1796

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -6,14 +6,21 @@ package examples
 
 import (
 	"bytes"
+	"context"
+	"encoding/json"
 	"path/filepath"
 	"testing"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ssm"
 	"github.com/pulumi/providertest/pulumitest"
 	"github.com/pulumi/providertest/pulumitest/assertpreview"
+	"github.com/pulumi/providertest/pulumitest/changesummary"
 	"github.com/pulumi/providertest/pulumitest/opttest"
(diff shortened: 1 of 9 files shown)


flostadler (approved):
Nice!

PR #1898, opened 2024-12-10
chore: add readOnlyProperties to metadata

corymhall:
This PR adds `readOnlyProperties` to the pulumi metadata. It also fixes
an issue with all properties (`readOnly/createOnly/writeOnly`) where
nested properties where not converted to the correct sdk name value.

### DIFF
--- a/provider/pkg/metadata/metadata.go
+++ b/provider/pkg/metadata/metadata.go
@@ -22,6 +22,7 @@ type CloudAPIResource struct {
 	AutoNamingSpec    *AutoNamingSpec                 `json:"autoNamingSpec,omitempty"`
 	Required          []string                        `json:"required,omitempty"`
 	CreateOnly        []string                        `json:"createOnly,omitempty"`
+	ReadOnly          []string                        `json:"readOnly,omitempty"`
 	WriteOnly         []string                        `json:"writeOnly,omitempty"`
 	IrreversibleNames map[string]string               `json:"irreversibleNames,omitempty"`
 	TagsProperty      string                          `json:"tagsProperty,omitempty"`
--- a/provider/pkg/metadata/metadata.go
+++ b/provider/pkg/metadata/metadata.go
@@ -16,17 +16,30 @@ type CloudAPIMetadata struct {
 
 // CloudAPIResource contains metadata for a single AWS Resource.
 type CloudAPIResource struct {
-	CfType            string                          `json:"cf"`
-	Inputs            map[string]pschema.PropertySpec `json:"inputs"`
-	Outputs           map[string]pschema.PropertySpec `json:"outputs"`
-	AutoNamingSpec    *AutoNamingSpec                 `json:"autoNamingSpec,omitempty"`
-	Required          []string                        `json:"required,omitempty"`
-	CreateOnly        []string                        `json:"createOnly,omitempty"`
-	ReadOnly          []string                        `json:"readOnly,omitempty"`
-	WriteOnly         []string                        `json:"writeOnly,omitempty"`
-	IrreversibleNames map[string]string               `json:"irreversibleNames,omitempty"`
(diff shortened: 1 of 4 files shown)


corymhall:

- #1898 ◀

provider/pkg/schema/resource_props_test.go
```diff
@@ -0,0 +1,36 @@
...
+		"nestedObject": {
+			input:    "Foo/BarBaz",
+			expected: "foo/barBaz",
+		},
+		"nestedArray": {
+			input:    "Foo/*/BarBaz",
+			expected: "foo/*/barBaz",
+		},
+		"nestedArrayObject": {
+			input:    "Foo/*/Bar/Baz",
```
flostadler, line 23: nit: Might be good to add multiple levels of array nesting to the test

provider/pkg/schema/resource_props.go
```diff
@@ -0,0 +1,31 @@
...
+// ToSdkName converts the property name to an SDK-compatible name.
+// This will also handle converting all nested properties.
+//   - Nested object properties are delimited with `/`
+//   - Nested array properties are delimited with `/*/`
+//
+// e.g.
+//   - `Foo/Bar/Baz` => `foo/bar/baz`
+//   - `Foo/*/BarBaz` => `foo/*/barBaz`
+func (r *ResourceProperty) ToSdkName() string {
+	arrayProps := strings.Split(string(*r), "/*/")
```
flostadler, line 22: I think this would panic right now when called on an `r` that's nil.

provider/pkg/metadata/metadata.go
```diff
@@ -22,6 +22,7 @@ type CloudAPIResource struct {
 	AutoNamingSpec    *AutoNamingSpec                 `json:"autoNamingSpec,omitempty"`
 	Required          []string                        `json:"required,omitempty"`
 	CreateOnly        []string                        `json:"createOnly,omitempty"`
+	ReadOnly          []string                        `json:"readOnly,omitempty"`
```
t0yv0, line 25: Here is an opportunity for a documentation comment on what this is about and a few examples of how this data looks like, specifically that nested names will be separated by "/", and that the names themselves are using Pulumi notation e.g. groupArn. If array element separator is used (is it '*')? also good to document there. It is still a little weird that "/" is the separator but that we can live with. 

Having a doc comment will save future maintainers some jumping around! Thank you.

provider/pkg/schema/resource_props.go
t0yv0: `contract.Assertf(r != nil, "ResourceProperty should not be nil")`

panics are good sometimes - this should not be nil I think.

t0yv0 (approved):

provider/pkg/schema/resource_props.go
flostadler: I'd argue that the provider shouldn't crash on a panic and rather exit with a proper error message.
If this shouldn't be nil, then we shouldn't accept a nil as the input imo

t0yv0: That's a better idea. I checked around and this seems fine:

```func (r ResourceProperty) ToSdkName() string```

Comment (t0yv0) [reply provider/pkg/schema/resource_props.go:22]: If you want to be more pedantic, which is not always bad, then this:

```
n := ResourceProperty(strings.TrimPrefix(v.(string), "/properties/"))
```

Should instead be:

```
parseResourceProperty(string) (ResourceProperty, error)
```

provider/pkg/metadata/metadata.go
corymhall: Added some docs.

provider/pkg/schema/resource_props.go
corymhall: Switched it to not be a pointer receiver.

flostadler (approved):

PR #1919, opened 2024-12-18
Implement autonaming configuration protocol

mikhailshilkov:
This PR implement the AWS Native part of https://github.com/pulumi/pulumi/issues/1518. See https://github.com/pulumi/pulumi/discussions/17592 for the full design.

In short, https://github.com/pulumi/pulumi/pull/17810 introduced the protobuf changes required for the provider-side implementation. With those, a provider can:

- Declare that it supports autonaming configurations with a response flag in Configure
- Accept two extra properties in CheckRequest: a proposed name and a mode to apply it

This PR applies those parameters to auto-name calculation if they are specified:

- In Disabled mode, we don't calculate auto-names and let the resource validation fail if an explicit name is not specified. For that, I had to thread an extra `bool` flag through autonaming machinery.
- In Enforce mode, we use whatever ProposedName is supplied by the engine
- In Propose mode, we take the ProposedName but can also apply trivia (e.g. add a suffix, as in the e2e test), or check the max/min length

I added unit tests to cover most autonaming cases. Also, added an end-to-end test in YAML. Apparently, end-2-end tests had no AWS credentials configured, so I added those. Let me know if that's against the intention.

Resolves https://github.com/pulumi/pulumi-aws-native/issues/1901

### DIFF
--- a/provider/pkg/autonaming/application.go
+++ b/provider/pkg/autonaming/application.go
@@ -15,10 +15,26 @@ type AutoNamingConfig struct {
 	RandomSuffixMinLength int  `json:"randomSuffixMinLength"`
 }
 
+// EngineAutonamingConfiguration contains autonaming parameters passed to the provider from the engine.
+type EngineAutonamingConfiguration struct {
+	RandomSeed     []byte
+	AutonamingMode *EngineAutonamingMode
+	ProposedName   string
+}
+
+// EngineAutonamingMode is the mode of autonaming to apply to the resource.
+type EngineAutonamingMode int32
+
+const (
+	EngineAutonamingModePropose EngineAutonamingMode = iota
+	EngineAutonamingModeEnforce
+	EngineAutonamingModeDisable
+)
+
 func ApplyAutoNaming(
 	spec *metadata.AutoNamingSpec,
 	urn resource.URN,
-	randomSeed []byte,
+	engineAutonaming EngineAutonamingConfiguration,
 	olds,
 	news resource.PropertyMap,
 	config *AutoNamingConfig,
@@ -27,11 +43,13 @@ func ApplyAutoNaming(
 		return nil
 	}
 	// Auto-name fields if not already specified
-	val, err := getDefaultName(randomSeed, urn, spec, olds, news, config)
+	val, ok, err := getDefaultName(urn, engineAutonaming, spec, olds, news, config)
 	if err != nil {
 		return err
 	}
-	news[resource.PropertyKey(spec.SdkName)] = val
+	if ok {
+		news[resource.PropertyKey(spec.SdkName)] = *val
+	}
 	return nil
 }
 
@@ -41,29 +59,58 @@ func ApplyAutoNaming(
 // based on its URN name, It ensures the name meets the length constraints, if known.
 // Defaults to the name followed by 7 random hex characters separated by a '-'.
 func getDefaultName(
-	randomSeed []byte,
 	urn resource.URN,
+	engineAutonaming EngineAutonamingConfiguration,
 	autoNamingSpec *metadata.AutoNamingSpec,
 	olds,
 	news resource.PropertyMap,
 	config *AutoNamingConfig,
-) (resource.PropertyValue, error) {
+) (*resource.PropertyValue, bool, error) {
 	sdkName := autoNamingSpec.SdkName
 
 	// Prefer explicitly specified name
 	if v, ok := news[resource.PropertyKey(sdkName)]; ok {
-		return v, nil
+		return &v, true, nil
 	}
 
 	// Fallback to previous name if specified/set.
 	if v, ok := olds[resource.PropertyKey(sdkName)]; ok {
-		return v, nil
+		return &v, true, nil
 	}
 
 	// Generate naming trivia for the resource.
 	namingTriviaApplies, namingTrivia, err := CheckNamingTrivia(sdkName, news, autoNamingSpec.TriviaSpec)
 	if err != nil {
-		return resource.PropertyValue{}, err
+		return nil, false, err
+	}
+
+	if engineAutonaming.AutonamingMode != nil {
+		//panic("yes")
+		switch *engineAutonaming.AutonamingMode {
+		case EngineAutonamingModeDisable:
+			return nil, false, nil
+		case EngineAutonamingModeEnforce:
+			v := resource.NewStringProperty(engineAutonaming.ProposedName)
+			return &v, true, nil
+		case EngineAutonamingModePropose:
+			proposedName := engineAutonaming.ProposedName
+
+			// Apply naming trivia to the generated name.
+			if namingTriviaApplies {
+				proposedName = ApplyTrivia(namingTrivia, proposedName)
+			}
+
+			// Validate the proposed name against the length constraints.
+			if autoNamingSpec.MaxLength > 0 && len(proposedName) > autoNamingSpec.MaxLength {
+				return nil, false, fmt.Errorf("proposed name %q exceeds max length of %d", proposedName, autoNamingSpec.MaxLength)
+			}
+			if autoNamingSpec.MinLength > 0 && len(proposedName) < autoNamingSpec.MinLength {
+				return nil, false, fmt.Errorf("proposed name %q is shorter than min length of %d", proposedName, autoNamingSpec.MinLength)
+			}
+
+			v := resource.NewStringProperty(proposedName)
+			return &v, true, nil
+		}
 	}
 
 	var autoTrim bool
@@ -94,7 +141,7 @@ func getDefaultName(
 		if left <= 0 && autoTrim {
 			autoTrimMaxLength := autoNamingSpec.MaxLength - namingTrivia.Length() - randomSuffixMinLength
 			if autoTrimMaxLength <= 0 {
-				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
+				return nil, false, fmt.Errorf("failed to auto-generate value for %[1]q."+
 					" Prefix: %[2]q is too large to fix max length constraint of %[3]d"+
 					" with required suffix length %[4]d. Please provide a value for %[1]q",
 					sdkName, prefix, autoNamingSpec.MaxLength, randomSuffixMinLength)
@@ -106,12 +153,12 @@ func getDefaultName(
 
 		if left <= 0 {
 			if namingTrivia.Length() > 0 {
-				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
+				return nil, false, fmt.Errorf("failed to auto-generate value for %[1]q."+
 					" Prefix: %[2]q is too large to fix max length constraint of %[3]d"+
 					" with required suffix %[4]q. Please provide a value for %[1]q",
 					sdkName, prefix, autoNamingSpec.MaxLength, namingTrivia.Suffix)
 			} else {
-				return resource.PropertyValue{}, fmt.Errorf("failed to auto-generate value for %[1]q."+
+				return nil, false, fmt.Errorf("failed to auto-generate value for %[1]q."+
 					" Prefix: %[2]q is too large to fix max length constraint of %[3]d. Please provide a value for %[1]q",
 					sdkName, prefix, autoNamingSpec.MaxLength)
 			}
@@ -123,9 +170,9 @@ func getDefaultName(
 	}
 
 	// Resource name is URN name + "-" + random suffix.
-	random, err := resource.NewUniqueName(randomSeed, prefix, randLength, maxLength, nil)
+	random, err := resource.NewUniqueName(engineAutonaming.RandomSeed, prefix, randLength, maxLength, nil)
 	if err != nil {
-		return resource.PropertyValue{}, err
+		return nil, false, err
 	}
 
 	// Apply naming trivia to the generated name.
@@ -133,7 +180,8 @@ func getDefaultName(
 		random = ApplyTrivia(namingTrivia, random)
 	}
 
-	return resource.NewStringProperty(random), nil
+	v := resource.NewStringProperty(random)
+	return &v, true, nil
 }
 
 // trimName will trim the prefix to fit within the max length constraint.
--- a/provider/pkg/autonaming/application.go
+++ b/provider/pkg/autonaming/application.go
@@ -43,11 +43,11 @@ func ApplyAutoNaming(
 		return nil
 	}
 	// Auto-name fields if not already specified
-	val, ok, err := getDefaultName(urn, engineAutonaming, spec, olds, news, config)
+	val, err := getDefaultName(urn, engineAutonaming, spec, olds, news, config)
 	if err != nil {
 		return err
 	}
-	if ok {
+	if val != nil {
 		news[resource.PropertyKey(spec.SdkName)] = *val
 	}
 	return nil
@@ -65,33 +65,33 @@ func getDefaultName(
 	olds,
 	news resource.PropertyMap,
 	config *AutoNamingConfig,
-) (*resource.PropertyValue, bool, error) {
+) (*resource.PropertyValue, error) {
 	sdkName := autoNamingSpec.SdkName
 
 	// Prefer explicitly specified name
 	if v, ok := news[resource.PropertyKey(sdkName)]; ok {
-		return &v, true, nil
+		return &v, nil
 	}
 
 	// Fallback to previous name if specified/set.
 	if v, ok := olds[resource.PropertyKey(sdkName)]; ok {
-		return &v, true, nil
+		return &v, nil
 	}
 
 	// Generate naming trivia for the resource.
 	namingTriviaApplies, namingTrivia, err := CheckNamingTrivia(sdkName, news, autoNamingSpec.TriviaSpec)
 	if err != nil {
-		return nil, false, err
+		return nil, err
 	}
 
 	if engineAutonaming.AutonamingMode != nil {
 		//panic("yes")
 		switch *engineAutonaming.AutonamingMode {
 		case EngineAutonamingModeDisable:
-			return nil, false, nil
+			return nil, nil
 		case EngineAutonamingModeEnforce:
 			v := resource.NewStringProperty(engineAutonaming.ProposedName)
-			return &v, true, nil
+			return &v, nil
 		case EngineAutonamingModePropose:
 			proposedName := engineAutonaming.ProposedName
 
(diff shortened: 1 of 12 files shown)


provider/pkg/autonaming/application.go
```diff
@@ -41,29 +59,58 @@ func ApplyAutoNaming(
...
+		return nil, false, err
+	}
+
+	if engineAutonaming.AutonamingMode != nil {
+		//panic("yes")
+		switch *engineAutonaming.AutonamingMode {
+		case EngineAutonamingModeDisable:
+			return nil, false, nil
+		case EngineAutonamingModeEnforce:
+			v := resource.NewStringProperty(engineAutonaming.ProposedName)
```
corymhall, line 93: Why don't we also validate the proposed name against the `autoNamingSpec`? I'm assuming that in the current case if the `engineAutonaming.ProposedName > autoNamingSpec.MaxLength` then the error will be thrown by the AWS service?

Same question for the naming trivia. If that is a requirement should we error here early?

provider/pkg/autonaming/application.go
```diff
@@ -41,29 +59,58 @@ func ApplyAutoNaming(
...
+			if autoNamingSpec.MinLength > 0 && len(proposedName) < autoNamingSpec.MinLength {
+				return nil, false, fmt.Errorf("proposed name %q is shorter than min length of %d", proposedName, autoNamingSpec.MinLength)
+			}
+
+			v := resource.NewStringProperty(proposedName)
+			return &v, true, nil
+		}
 	}
 
 	var autoTrim bool
```
corymhall, line 116: Is there a way that we can get the new engine autonaming to work with the existing autoNaming config? It seems like it could work with `EngineAutonamingModePropose`. And then maybe we should throw errors if you try to configure it with `Enforce` or `Disable`?

https://github.com/pulumi/pulumi-aws-native/blob/11ac7b3942424e3c4863677b2a6ead0de2ab37bc/provider/pkg/schema/gen.go#L188-L191

corymhall (changes requested):

provider/pkg/autonaming/application.go
mikhailshilkov: That's the intended semantics of enforce, as described in the spec. It can be used to overcome an erroneous autonaming configuration of the provider and override it to follow through to the AWS API.

provider/pkg/autonaming/application.go
mikhailshilkov: My original take was a no, but now I think I misunderstood `autoTrim`. It could potentially apply to the Propose mode. `randomSuffixMinLength` won't work because the suffix (or prefix, or whatever format) is already baked into the proposed name.

mikhailshilkov: Actually, I'm still not sure about `autoTrim`. Do we have any docs about what its supposed to do exactly? Reading the code, its semantics is roughly "if a random suffix makes the generated name too long, we can trim that random suffix down to N characters, where N is either 1 or whatever the user configured". I don't think there is any way to apply that to the proposed name, given it has an unknown structure. I assume we don't want to be trimming non-random portions of the name.

.github/workflows/run-acceptance-tests.yml
```diff
@@ -140,6 +140,15 @@ jobs:
       with:
         name: pulumi-${{ env.PROVIDER }}-provider.tar.gz
         path: ${{ github.workspace }}/bin/provider.tar.gz
+    - name: Configure AWS Credentials
```
flostadler, line 143: This file is auto generated, we'll need to change it here: https://github.com/pulumi/ci-mgmt

provider/pkg/autonaming/application.go
```diff
@@ -15,10 +15,26 @@ type AutoNamingConfig struct {
 	RandomSuffixMinLength int  `json:"randomSuffixMinLength"`
 }
 
+// EngineAutonamingConfiguration contains autonaming parameters passed to the provider from the engine.
+type EngineAutonamingConfiguration struct {
```
flostadler, line 19: Curious, why are we not reusing the autonaming structs and enum from pu-sdk?

provider/pkg/autonaming/application.go
```diff
@@ -41,29 +59,57 @@ func ApplyAutoNaming(
...
 func getDefaultName(
-	randomSeed []byte,
 	urn resource.URN,
+	engineAutonaming EngineAutonamingConfiguration,
 	autoNamingSpec *metadata.AutoNamingSpec,
 	olds,
 	news resource.PropertyMap,
 	config *AutoNamingConfig,
-) (resource.PropertyValue, error) {
+) (*resource.PropertyValue, bool, error) {
```
flostadler, line 68: Nit: Is the boolean just used to indicate whether a name is returned? I feel like we could omit that because we should non-nil/non-empty check the name to ensure a bug in autonaming is not silently setting a name to the empty string. (i.e. if no error is returned but the name is empty we know something is fishy)

Just a nit, feel free to ignore

provider/pkg/autonaming/application.go
flostadler: I think this is gonna be tricky for users right now. After reading https://github.com/pulumi/pulumi/discussions/17592 I couldn't find a way users could configure to shorten a logical name. IMO one of the reasons users would want to configure the autonaming behavior in aws-native is to get around naming constraints that are missing from the CFN schema. Usually that's a missing maxLength constraint.

A substring expression would be great for autonaming. Something like this maybe?`${substr(name, 56)}-${alphanum(7)}`

provider/pkg/autonaming/application.go
mikhailshilkov: Great question. I started there but then went for keeping the `autonaming` module self-contained. It feels like a fairly complicated piece of "domain logic", so I wasn't sure I wanted to mix in RPC types. I'm not sure how much I bought by doing so - can be persuaded the other way.

provider/pkg/autonaming/application.go
mikhailshilkov: Yeah, they can't shorten a logical name right now. I haven't seen this concern anywhere in the original issue or RFC discussion, but if it comes up often enough, we can add it later. The good news is that it will then work for all providers equally, not just AWS Native.

provider/pkg/autonaming/application.go
mikhailshilkov: I'm probably overthinking it indeed. Pushed a commit to simplify back to two return values and nil checks.

.github/workflows/run-acceptance-tests.yml
mikhailshilkov: I can do this - I guess my first question is whether I _should_ be configuring credentials here and why I'm the first one to need this. Should I be using some other place for end-2-end tests?

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -30,6 +31,19 @@ func TestE2eSnapshots(t *testing.T) {
 	})
 }
 
+func TestAutonaming(t *testing.T) {
```
flostadler, line 34: we should skip this test when running in short mode (i.e. unit tests only)

.github/workflows/run-acceptance-tests.yml
flostadler: I think it makes sense to do that for now. They already include upgrade tests and while they do not use any invokes right now, they might in the future and those need creds: https://github.com/pulumi/pulumi-aws-native/blob/11ac7b3942424e3c4863677b2a6ead0de2ab37bc/provider/pkg/provider/provider_2e2_test.go#L19-L31

(Just saw that the upgrade tests are missing `-short` flag, I'll add that separately)

Long term I'd like us to have a more standard test setup where we have a single go.mod on the root and then separate directories for tests in order to keep slower-running integration tests separate from unit tests. This would enable us to run unit tests without needing to remember `-short` or build tags. But that's all unrelated to your changes.

provider/pkg/autonaming/application.go
corymhall: Without/pre`autoTrim` the behavior was to allow the random suffix to be as little as 1 character if that allowed the name to fit within the maxLength requirements.

`autoTrim` introduced a couple of new behaviors.

1. `randomSuffixMinLength` could be used to enforce a minimum length of the random suffix
2. `autoTrim: true`. If the length of the name exceeded the maxLength (including the random suffix) then the name would be trimmed to fix within the maxLength. The part of the name that is trimmed is taken from the middle of the name in order to keep uniqueness. 

It sounds like the `randomSuffixMinLength` might conflict with the new engine functionality. In that case we should throw an error if both are used. Since `autoTrim` removes from the middle it's possible it could work with the new engine behavior, but I think it would also be fine if these two features are mutually exclusive for now (especially if the autotrim behavior could be added to the engine)

provider/pkg/autonaming/application.go
corymhall: > . It can be used to overcome an erroneous autonaming configuration of the provider and override it to follow through to the AWS API.

Ok so it's like an escape hatch that users can use until the upstream bug is fixed?

mikhailshilkov: It is an escape hatch too. More broadly, a lot of users expressed strong opinions that they want Pulumi to stop mangling with names, period. Now they can do that easily and we guarantee that users have full control.

provider/pkg/autonaming/application.go
flostadler: +1 to what Cory said.
Are we able to make a distinction between whether engine level autonaming is configured for the stack/provider or only for a certain resource.

I feel like `autoTrim` + stack/provider level autonaming config should be an invalid case, but `autoTrim` + resource level autonaming config should be valid because it is essentially an escape hatch.

mikhailshilkov: > Are we able to make a distinction between whether engine level autonaming is configured for the stack/provider or only for a certain resource?

No

> I feel like autoTrim + stack/provider level autonaming config should be an invalid case, but autoTrim + resource level autonaming config should be valid because it is essentially an escape hatch.

Could you please explain why this distinction? If a user configures a pattern like `${stack}-${name}-${hex(6)}` on the provider level for all AWS Native resources vs. does so for a give `aws-native:foo:Bar` - how does that change the expectation of the user of whether some middle part of the generated name will be trimmed or not?

In any case, the question is a bit hypothetical because the provider can't tell the difference with the current design. I think I'm inclined to do the following for `Propose` mode:

1. If `RandomSuffixMinLength` is configured, throw an error, we can't do anything with it.
2. If `AutoTrim` is true and `autoNamingSpec.MaxLength` is defined, call `trimName(ProposedName, autoNamingSpec.MaxLength)` and hope it does what the user wanted.

But we can also keep it simple and throw an error for now.

Thoughts?

.github/workflows/run-acceptance-tests.yml
mikhailshilkov: Working on CI in https://github.com/pulumi/ci-mgmt/pull/1239 but also discussing offline what best approach could be

blampe: https://github.com/pulumi/ci-mgmt/pull/1239 is typically how we'd unblock this, but we're trying to do a couple things right now in the opposite direction:
- we want to reduce the number of exceptions for native providers in order to make it easier to consolidate them with bridged workflows
- and we want to make CI tests more reproducible locally, which means reducing the reliance on CI-specific setup steps.

For those reasons I'll suggest an alternative approach. Basically just perform the auth in-process as part of your test's setup. Actually, before you do that maybe just plumb the environment variables through, since the provider should login for you? So the test would get an environmnet

```go
Env: []string{"AWS_ROLE_SESSION_NAME=" + os.Getenv("PROVIDER"), "AWS_ROLE_TO_ASSUME=" + os.Getenv("AWS_CI_ROLE_ARN"), ...}
```
Otherwise, if you do need to perform the auth, something like this pseudocode:

```go
cfg := aws.Config{
  Credentials: credentials.NewStaticCredentialsProvider(accessKey, secretKey, ""),
  Region: region,
}

c := sts.NewFromConfig(cfg)
creds, err := c.AssumeRole(ctx, &sts.AssumeRoleInput{
  RoleArn:         aws.String(os.Getenv("AWS_CI_ROLE_ARN")),
  RoleSessionName: aws.String(os.Getenv("AWS_ROLE_SESSION_NAME")),
})
require.NoError(t, err)

// Set environment with creds
```


Then pass those environment variables to the e2e test via `Env:`, or (worse) if you're lazy use `os.Setenv`.

If `provider/pkg/provider/provider_2e2_test.go` is the only place where you need these credentials, then I suggest defining a helper there like `loginToAWS(t)`. You can put it behind a [`sync.Once`](https://pkg.go.dev/sync) to prevent logging in more than you need to.

Give it a try, and if it turns into a rabbit hole we can merge the ci-mgmt PR and clean it up later.

mikhailshilkov: I can give this a try... This would make it unusable in a local dev environment unless one replicates all env vars, right?

provider/pkg/autonaming/application.go
corymhall: I think it's fine for now to just say that you can't configure both the engine autonaming _and_ the provider autoName config and throw an error/warning if you have explicitly configured both.

I can create an issue for supporting a `autoTrim` type feature in the engine, would that be in pu/pu?

.github/workflows/run-acceptance-tests.yml
flostadler: I'd personally vote for the `ci-mgmt` option because it doesn't mess with the local dev setup. The changes to `ci-mgmt` are not that big IMO and the aws-native test structure will have to change for the consolidation anyways

provider/pkg/autonaming/application.go
mikhailshilkov: Added returning an error if both engine and provider configs are specified. Also, renamed a few types and variables to make it consistent and more clear.

> I can create an issue for supporting a autoTrim type feature in the engine, would that be in pu/pu?

yes

.github/workflows/run-acceptance-tests.yml
mikhailshilkov: I think I prefer `ci-mgmt` too... I un-drafted that PR, happy to ship it and remove the CI change here after it percolates.

blampe: > This would make it unusable in a local dev environment unless one replicates all env vars, right?

No... your local credentials are used to assume the CI role. This makes things _more_ consistent between local and CI.

> I'd personally vote for the ci-mgmt option because it doesn't mess with the local dev setup.

Nothing changes with local setup. The test would just automatically assume the CI role using your local credentials.

For completeness a more complete example of what I mean is below. This gives me an "SSO session has expired or is invalid" error so I'm definitely doing something wrong but you get the idea.

Looking at the CI role though it I think the real blocker is that only root is allowed to assume it. https://github.com/pulumi/ci-mgmt/issues/1262

```diff

 func TestAutonaming(t *testing.T) {
        skipIfShort(t)
-       pt := newAwsTest(t, filepath.Join("testdata", "autonaming"), opttest.Env("PULUMI_EXPERIMENTAL", "1"))
+
+       if os.Getenv("AWS_ACCESS_KEY_ID") == "" {
+               t.Skip("missing AWS_ACCESS_KEY_ID")
+       }
+       if os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
+               t.Skip("missing AWS_SECRET_ACCESS_KEY")
+       }
+       if os.Getenv("AWS_REGION") == "" {
+               t.Skip("missing AWS_REGION")
+       }
+
+       region := os.Getenv("AWS_REGION")
+
+       initialCredentials := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
+               os.Getenv("AWS_ACCESS_KEY_ID"),
+               os.Getenv("AWS_SECRET_ACCESS_KEY"),
+               os.Getenv("AWS_SESSION_TOKEN"),
+       ))
+
+       cfg, err := config.LoadDefaultConfig(context.Background(),
+               config.WithRegion(region),
+               config.WithCredentialsProvider(initialCredentials),
+       )
+       require.NoError(t, err)
+
+       // Assume CI role.
+       stsClient := sts.NewFromConfig(cfg)
+       roleCredentials := aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(stsClient, "arn:aws:iam::894850187425:role/ContinuousDeliveryAdminRole", func(opts *stscreds.AssumeRoleOptions) {
+               opts.RoleSessionName = "aws-native@githubActions"
+               opts.Duration = 1 * time.Hour
+       }))
+       assumedCreds, err := roleCredentials.Retrieve(context.Background())
+       env := []opttest.Option{
+               opttest.Env("PULUMI_EXPERIMENTAL", "1"),
+               opttest.Env("AWS_ACCESS_KEY_ID", assumedCreds.AccessKeyID),
+               opttest.Env("AWS_SECRET_ACCESS_KEY", assumedCreds.SecretAccessKey),
+               opttest.Env("AWS_SESSION_TOKEN", assumedCreds.SessionToken),
+               opttest.Env("AWS_ACCOUNT_ID", "894850187425"),
+               opttest.Env("AWS_REGION", region),
+               opttest.Env("AWS_PROFILE", ""),
+       }
+
+       pt := newAwsTest(t, filepath.Join("testdata", "autonaming"), env...)
```

flostadler: @blampe I don't think the local admin role is allowed to assume the CI role right now: https://github.com/pulumi/home/blob/6d98acbc91c4cc4deea122a18cb0503b41354b3a/aws/accounts/pulumi-ci.ts#L39

blampe: Exactly https://github.com/pulumi/ci-mgmt/issues/1262 :)

provider/pkg/autonaming/application.go
flostadler: > Could you please explain why this distinction? If a user configures a pattern like ${stack}-${name}-${hex(6)} on the provider level for all AWS Native resources vs. does so for a give aws-native:foo:Bar - how does that change the expectation of the user of whether some middle part of the generated name will be trimmed or not?

My reasoning is that users might want to apply the trimming by default to ensure their resources fit AWS limitations, but then overwrite it for a particular resource type as an escape hatch (e.g. Bucket only allowing lowercase chars).
But it's fine for now because the provider can't detect this atm.

provider/pkg/autonaming/application.go
flostadler: I don't have a strong preference either way, was mostly curious. I like decoupling it from RPC types!

flostadler (approved):
LGTM!

corymhall (approved):

PR #2003, opened 2025-01-21
Remove unsupported LangVersion from dotnet example

flostadler:
dotnet builds started failing, due to the upgrade of the github runners to ubuntu 24.04.
The `RoleCS` test was failing with issues in selecting `LanguageVersion` 11.

I've removed that version identifier from the `csproj` of the failing test to align with the other CS tests and changed it to not use features from language version 11.

### Why did this work before?
Previously this worked, because the test implicitly used v8 of the dotnet runtime instead of the requested v6 (configured in ci-mgmt https://github.com/pulumi/ci-mgmt/issues/1307).

The `setup-dotnet` action implicitly installs the latest LTS before installing other versions.
We're requesting 6.0.x and on Ubuntu 22.04 this was already pre-installed, so this happened:
- Install latest LTS (v8)
- Install v6.0.x -> Already present -> Noop

Because of that v8 was on the path first and used for tests.

Now on Ubuntu 24.04 v8 is pre-installed, so this happens:
- Install latest LTS (v8) -> Already present -> Noop
- Install v6.0.x

Now we're actually using the requested v6, but the `RoleCs` test is not compatible with that.

Fixes https://github.com/pulumi/pulumi-aws-native/issues/2004


examples/role-cs/role-cs.csproj
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>11</LangVersion>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
```
thomas11, line 5: You probably want to bump this to `net8.0` as well. We saw [jobs fail in azure native due to the removal of 6.0 when runner images were updated](https://github.com/pulumi/pulumi-azure-native/pull/3868).

t0yv0 (approved):

examples/role-cs/role-cs.csproj
flostadler: I tried that first, but turns out there's a bigger issue with most of the native providers: https://github.com/pulumi/ci-mgmt/issues/1307
The installation is rather broken right now, before github upgraded the runners it installed `8.0` and then `3.1.301`, right now it's installing `8.0`, `6.0` then `3.1.301`.

Somehow `6.0` is on the path before `8.0` now and that does not support `LangVersion` 11.

flostadler: We'll first need to fix https://github.com/pulumi/ci-mgmt/issues/1307 before being able to upgrade the target framework

PR #2118, opened 2025-03-10
SDK generation for latest CFN schemas

### LINKED ISSUE #2109: Workflow failure: nightly-sdk-generation
## Workflow Failure

[nightly-sdk-generation](https://github.com/pulumi/pulumi-aws-native/blob/master/.github/workflows/nightly-sdk-generation.yml) has failed. See the list of failures below:

- [2025-03-07T04:41:58.000Z](https://github.com/pulumi/pulumi-aws-native/actions/runs/13713885048)

- [2025-03-10T04:40:06.000Z](https://github.com/pulumi/pulumi-aws-native/actions/runs/13756970395)

- [2025-03-11T04:41:42.000Z](https://github.com/pulumi/pulumi-aws-native/actions/runs/13780487431)

corymhall:
This updates the SDK based on the latest CFN schemas. I've also included
and update that preserves resources that have been completely removed
from the CFN schema.

Typically the deprecated-types.txt contains
resources that had CCAPI enabled and then CCAPI was later disabled. We
never handled the case where resources were completely removed from the
CFN schema, but AWS has started to do so now that they are deprecating
services. I've added logic that keeps deprecated schemas instead of
removing them.

The first two commits have my code changes, the rest have generated code

fixes #2109


provider/cmd/pulumi-gen-aws-native/main.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
+	"slices"
```
t0yv0, line 23: Lint would put it next to "strings" I think. Maybe lint is not setup here.

provider/pkg/schema/gen.go
```diff
@@ -1482,9 +1482,8 @@ func (ctx *cfSchemaContext) genEnumType(enumName string, propSchema *jsschema.Sc
 			continue
 		}
 
-		// Special case for `ChannelPreset` enum in the `AWS::IVS::Channel` resource. The enum has an empty string value which
-		// gets returned by the service in certain default cases (for channel types (BASIC and STANDARD)).
-		if typName == "ChannelPreset" && str == "" {
+		// Special case for when there is an enum which has an empty string value
```
t0yv0, line 1485: Is there a linked TODO?

t0yv0 (approved):

provider/pkg/schema/gen.go
corymhall: No, but good call. We should automate this case.

PR #2480, opened 2025-09-30
Replace pulumi-java-gen with `pulumi package gen-sdk`

mjeffryes:
One step closer to [Remove pulumi-java-gen · Issue #1636 · pulumi/ci-mgmt](https://github.com/pulumi/ci-mgmt/issues/1636)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,6 @@ CODEGEN         := pulumi-gen-${PACK}
 
 WORKING_DIR		:= $(shell pwd)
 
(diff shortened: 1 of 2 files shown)


guineveresaenger (approved):
....wait, this Just Works? :shipit:

guineveresaenger (changes requested):
Okay, I spoke too soon.

You'll also have to ignore the gradle build file when checking for the clean worktree, since the version will have a diff.

Issue #1983, opened 2025-01-13
Flaky test `TestAutonaming`


flostadler:
Got this error:
```
          aws-native:sqs:Queue (queue):
            error: creating resource: creating resource (await): operation CREATE failed with "Throttling": You must wait 60 seconds after deleting a queue before you can create another with the same name. (Service: Sqs, Status Code: 400, Request ID: 7503e9ea-2fa2-5fba-b1cc-b6117f9af4d4)
```

We might need some randomization for the name (or use another resource that doesn't have this limitation).

Issue #1959, opened 2025-01-02
Resource level retry logic


corymhall:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Each AWS service has different API limits and require different retry logic. The CCAPI schema does not contain this information so we currently have global retry logic which applies to all resources. 

For example the `` resource can have a large propagation delay when creating. The terraform-aws-provider has it configured to retry for up to 5 minutes. [hashicorp/terraform-provider-aws@7a8669e/internal/service/ec2/vpc_route_table_association.go?plain=1#L80](https://github.com/hashicorp/terraform-provider-aws/blob/7a8669e827e3fbe32f436e93ef77a4c58a199751/internal/service/ec2/vpc_route_table_association.go?plain=1#L80). It doesn't make sense to make 5 minutes the global retry limit though.

corymhall:
Another example https://github.com/pulumi/pulumi-cdk/issues/277 where ApiGateway can only delete 1 api per 30 seconds.

flostadler:
Another example is SQS queues: https://github.com/pulumi/pulumi-aws-native/issues/1983

You need to wait 1 minute after deleting a queue before creating one with the same name again

corymhall:
Maybe related to https://github.com/pulumi/pulumi/issues/7932

blampe:
I was sniped by this today and spent some time learning about how AWS configures retries.

In the API Gateway case it's returning a "Throttling" error with a 429 status code. That's not in the "Standard" collection of retryable errors, but it _is_ in the "Adaptive" set.

https://github.com/aws/aws-sdk-go-v2/blob/81724c7085d59bbb1c4f696f8b1d9c11b007942d/aws/retry/standard.go#L42-L92

The Gamelift API appears to have been updated to also return these as Throttling instead of GeneralServiceException errors. So, it _seems_ like there's an easy change to make use of Adaptive retries that would help eliminate a whole category of problems.

In the SQS "recently deleted" case it's returning a 400 status code which never gets retried. It is possible to mark errors as retry-able on a per-service basis, so this also _seems_ like it should be easy to fix.

I was able to reproduce these errors with Pulumi programs as well as via vanilla clients -- that tells me the provider itself isn't so misconfigured that it's causing these. 

[...]

Issue #1780, opened 2024-10-21
Throttling exceptions trying to provision GameLift Resources


automagic:
### What happened?

When creating GameLift resources, the provider receives throttling errors from the CloudControl API, but does not appear to handle these gracefully (backoff, etc), and fails to create resources.

Additionally the Pulumi output says "HandlerErrorCode: GeneralServiceException" as opposed to the AWS common error "ThrottlingException" when receiving these exceptions, making them hard to 

### Example

n/a

### Output of `pulumi about`

latest

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

automagic:
May be similar to https://github.com/pulumi/pulumi-aws-native/issues/1200

flostadler:
Thanks for raising this @automagic!

What's the indication for this being caused by throttling? `HandlerErrorCode: GeneralServiceException` is returned by AWS CloudControl if the service returns an error that cannot be mapped to any known error (`ThrottlingException` is one of those).

Does this happen for all GameLift resources or just specific ones?
Could you please provide a repro we can use to get to the bottom of this issue, thanks!

mjeffryes:
I'm assuming by the lack of updates that this is no longer a priority concern @automagic, but feel free to reopen if you'd like to dig in with us on a reproduction.

automagic:
Here is the reporoduction code:

```
using System.Collections.Generic;
using Pulumi;
using Pulumi.AwsNative.GameLift;
using MatchmakingRuleSet = Pulumi.AwsNative.GameLift.MatchmakingRuleSet;
using MatchmakingRuleSetArgs = Pulumi.AwsNative.GameLift.MatchmakingRuleSetArgs;
using MatchmakingConfiguration = Pulumi.AwsNative.GameLift.MatchmakingConfiguration;
using MatchmakingConfigurationArgs = Pulumi.AwsNative.GameLift.MatchmakingConfigurationArgs;

return await Deployment.RunAsync(() =>
{
    var westProvider = new Pulumi.AwsNative.Provider("west", new Pulumi.AwsNative.ProviderArgs
    {
        Region = "us-west-2",
    });

    var eastProvider = new Pulumi.AwsNative.Provider("east", new Pulumi.AwsNative.ProviderArgs
    {
        Region = "us-east-1",
    });

    MakeRegionalResources(westProvider, "west");
[...]

PR #2522, opened 2025-10-17
Handle Throttling responses with adaptive retries

blampe:
This fixes what I believe was a significant misconfiguration of the provider's retry logic.

In particular, we configure our AWS client for 25 retries but this has no bearing on downstream CloudControl requests. That is, we will retry when submitting a request to CC, but all of the actual work happens asynchronously using default retry settings, i.e. 3 retries + ignore throttling.

Documentation is sparse on CC's error handling, and I had mistakenly assumed it would inherit the same retry settings we configured our client with when interacting resources, but this is not the case.

As far as I can tell, the intent (e.g. https://github.com/pulumi/pulumi-aws-native/pull/862) was to have these retries apply to the underlying CC Create/Update/Delete operations, but these always succeed and return a progress handle. We then query that progress handle, and it eventually indicates the underlying operation failed with a "Throttling" error code. SDK retries don't kick in because all of _our_ client's requests -- kicking off the operation, and then querying for its progress -- were successful.

So CloudControl sees the errors and surfaces them to us as "FAILED" progress events, but we still want to retry some of the requests, especially Throttling. I don't want to introduce a bunch of bespoke retry logic when the AWS SDK already provides a ton of reasonable defaults, so let's try to leverage that.

This PR makes the following changes:
* We switch from the Standard to the [AdaptiveMode](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v1.39.2/aws/retry#AdaptiveMode) retryer. AdaptiveMode is essentially the same as Standard, but it will retry automatically on Throttling errors instead of raising them. This is important because all of our examples (https://github.com/pulumi/pulumi-aws-native/issues/1983, https://github.com/pulumi/pulumi-aws-native/issues/1780, https://github.com/pulumi/pulumi-cdk/issues/277) are examples of Throttling errors that our Standard retryer doesn't handle. 

  > AdaptiveMode provides an experimental retry strategy that expands on the Standard retry  strategy, adding client attempt rate limits. The attempt rate limit is initially unrestricted, but becomes restricted when the attempt fails with for a throttle error. When restricted AdaptiveMode may need to sleep before an attempt is made, if too many throttles have been received.
  >
  > Eventually unrestricted attempt rate limit will be restored once attempts no longer are failing due to throttle errors.

* We add a method `progressEventToError` which maps error codes on progress events to first-class error types. This allows us to use our existing `cfg.Retryer` (the one configured for 25 retries) with these events.

* We add a `withRetries` method which wraps a Create/Update/Delete (+await) operation. If during the await we see that the operation failed due to Throttling, it will be retried using exactly the same retry logic we've configured our AWS client with.

I was able to reproduce the SQS and GameLift errors and confirmed that the provider now works as expected with them. Here's an example of an SQS queue that was created right after being deleted:

```
❯ pulumi up --yes --skip-preview
Updating (dev):
     Type                     Name                 Status            Info
 +   pulumi:pulumi:Stack      aws-native-1959-dev  created (66s)     3 warnings
 +   └─ aws-native:sqs:Queue  foo                  created (65s)

Diagnostics:
  aws-native:sqs:Queue (foo):
```

Along the way the user sees status messages like this:

>    Retrying after "Throttling" error (attempt 1/25, waiting 1.579690526s)
>    Retrying after "Throttling" error (attempt 2/25, waiting 86.117105ms)
>    Retrying after "Throttling" error (attempt 3/25, waiting 4.734073457s)
>    Retrying after "Throttling" error (attempt 4/25, waiting 2.157481486s)
>    Retrying after "Throttling" error (attempt 5/25, waiting 20s)

Refs https://github.com/pulumi/pulumi-aws-native/issues/1959.

### DIFF
--- a/provider/pkg/client/client.go
+++ b/provider/pkg/client/client.go
@@ -281,7 +281,7 @@ func (c *clientImpl) withRetries(
 			return pi, err // Nothing else to do.
 		}
 
-		// Try to get a retry token (handles rate limiting)
+		// Try to get a retry token.
 		releaseRetryToken, tokenErr := retryer.GetRetryToken(ctx, sdkErr)
 		if tokenErr != nil {
 			if c.logger != nil {
@@ -292,7 +292,7 @@ func (c *clientImpl) withRetries(
 			return pi, err
 		}
 
-		// Check if we've exhausted retry attempts
+		// Check if we've exhausted retry attempts.
 		if attempt >= maxAttempts {
 			_ = releaseRetryToken(sdkErr)
 			if c.logger != nil {
@@ -301,9 +301,13 @@ func (c *clientImpl) withRetries(
 			return pi, err
 		}
 
-		// Calculate retry delay using the SDK's backoff strategy.
-		delay, delayErr := retryer.RetryDelay(attempt, sdkErr)
-		if delayErr != nil {
+		// If CloudControl gave us an explicit RetryAfter then use it,
+		// otherwise calculate delay using the SDK's backoff strategy.
+		var delay time.Duration
+		var delayErr error
+		if pi.RetryAfter != nil && time.Now().Before(*pi.RetryAfter) {
+			delay = time.Until(*pi.RetryAfter)
+		} else if delay, delayErr = retryer.RetryDelay(attempt, sdkErr); delayErr != nil {
 			_ = releaseRetryToken(sdkErr)
 			if c.logger != nil {
 				_ = c.logger.LogStatus(ctx, diag.Warning, urn, fmt.Sprintf(


blampe:

- #2530
    - #2522 ◀

provider/pkg/client/client.go
```diff
@@ -215,3 +222,142 @@ func (*clientImpl) getResourceRetryNotFoundRetrySettings() (int, *retry.Exponent
...
+			))
+		}
+
+		// Wait to retry.
+		select {
+		case <-ctx.Done():
+			_ = releaseToken(sdkErr)
+			return nil, ctx.Err()
+		case <-time.After(delay):
+			_ = releaseToken(nil) // nil indicates retry will proceed.
```
EronWright: I'm sure you'd like to have `releaseToken` be in a defer block.  Line 357 seems to be the only special case, so maybe you could just set `sdkErr = nil` here and do `defer { releaseToken(sdkErr) }`.

EronWright (commented):
Seems like a solid plan.

corymhall (commented):
This is great!

Do you know how long it would take for all 25 retries to be exhausted? We should probably document that as the global upper bound. It could potentially lead to cases where it takes us a _long_ time to eventually fail. The AWS provider has things like this to track the length of time each operation _may_ take https://github.com/hashicorp/terraform-provider-aws/blob/c99645b43d640948544b34a2f25fe3c967d81bc1/internal/service/ec2/wait.go#L23-L66. It looks like there can be a pretty big range which means we'll have tradeoffs in one direction or another (we'll either fail after 3 minutes when we should have waited 10 or we'll wait for 10 when we could have failed faster after 3).

provider/pkg/client/client.go
blampe: Yeah, normally I would prefer to defer this but we're in a loop, so the defer would only happen on function exit. The retryer uses a token bucket to throttle itself, so it's important to release the retry token as we go.

provider/pkg/client/client.go
```diff
@@ -73,15 +76,16 @@ func (c *clientImpl) Create(ctx context.Context, urn resource.URN, typeName stri
...
-		return nil, nil, fmt.Errorf("creating resource: %w", err)
-	}
-
-	pi, waitErr := c.awaiter.WaitForResourceOpCompletion(ctx, res)
+	pi, waitErr := c.withRetries(ctx, urn, func() (*types.ProgressEvent, error) {
+		res, err := c.api.CreateResource(ctx, typeName, payload)
+		if err != nil {
+			return nil, err
+		}
+		return c.awaiter.WaitForResourceOpCompletion(ctx, res)
```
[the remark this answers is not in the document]
blampe: Yeah I wasn't sure if reads would also be in scope for something like this. I noticed some similiar-ish retry logic for GetResource, which wants retries only in some situations. In any case I opted not to touch it since the other operations are giving us most of the trouble.

blampe:
> Do you know how long it would take for all 25 retries to be exhausted? We should probably document that as the global upper bound. It could potentially lead to cases where it takes us a long time to eventually fail. The AWS provider has things like this to track the length of time each operation may take https://github.com/hashicorp/terraform-provider-aws/blob/c99645b43d640948544b34a2f25fe3c967d81bc1/internal/service/ec2/wait.go#L23-L66. It looks like there can be a pretty big range which means we'll have tradeoffs in one direction or another (we'll either fail after 3 minutes when we should have waited 10 or we'll wait for 10 when we could have failed faster after 3).

The current config does some exponential jitter for a few retries and then maxes out at 20s per attempt, so we won't wait more than 500s in total. What do you think would be the best place to document the behavior?

It turns out CloudControl can also return a `RetryAfter` property on the progress event. I'm not sure under what conditions it's actually set, but I figure we should respect it if it's populated so I made that change. I would _expect_ `RetryAfter` to help with those very long-running operations, but it all depends on whether the server sends us that information so who knows.

provider/pkg/provider/provider_2e2_test.go
```diff
@@ -46,6 +48,21 @@ func TestAutonaming(t *testing.T) {
...
+	skipIfShort(t)
+
+	// Pick a slightly random name so as to not conflict with concurrent CI runs.
+	name := fmt.Sprintf("throttling-queue-%d", rand.Intn(1000))
+	test := newAwsTest(t, filepath.Join("testdata", "throttling"))
+	test.SetConfig(t, "queueName", name)
+	_ = test.Up(t)
+	_ = test.Destroy(t)
+	_ = test.Up(t)
+}
```
blampe: Fails on master.

PR #2530, opened 2025-10-20
Expose resource-scoped logging to the CloudControl client

blampe:
This plumbs the URN down to our CC client so we can log status messages
for resources. This will allow us to log when we are retrying operations
due to Throttling errors.


blampe:

- #2530 ◀
    - #2522

PR #2688, opened 2025-12-11
fix(efs): EFS Replication Refresh/Up

### LINKED ISSUE #2686: Handle `propertyTransform` in the schema as diff suppression functions
The CloudFormation schema has a `propertyTransform` object. For example:

```json
"propertyTransform" : {
    "/properties/FileSystemProtection/ReplicationOverwriteProtection" : "$uppercase(FileSystemProtection.ReplicationOverwriteProtection)='DISABLED' ? 'REPLICATING' : $uppercase(FileSystemProtection.ReplicationOverwriteProtection)",
  },
```

This is used in CloudFormation's drift detection feature, so it would look at the user's template, transform it to REPLICATING  before comparing it to the cloud value.

https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-model-false-drift.html

We should figure out a way of automatically handling this.

rshade:
The fix for the EFS replication refresh/up failure issue is complete and tested. Here's what was implemented:

Problem

When a user creates an EFS FileSystem with fileSystemProtection.replicationOverwriteProtection: "DISABLED" and it becomes a replication destination, AWS transitions this value to REPLICATING. After pulumi refresh followed by pulumi up, the provider would attempt to update this property back to DISABLED, causing AWS to reject the update.

Solution

Implemented filterAWSManagedPropertyDiffs in provider/pkg/provider/provider.go that:

- Checks specifically for fileSystemProtection.replicationOverwriteProtection == "REPLICATING"
- Suppresses the diff only when the value is REPLICATING (AWS-managed state)
- Does NOT mark the property as read-only (users can still set it during creation)

Files Changed

| File                                                                  | Description                                                           |
|-----------------------------------------------------------------------|-----------------------------------------------------------------------|
| provider/pkg/provider/provider.go                                     | Added filterAWSManagedPropertyDiffs function called in Diff()         |
| provider/pkg/schema/inputs.go                                         | Helper functions for filtering properties with nested read-only paths |
| provider/pkg/schema/inputs_test.go                                    | Unit tests for the helper functions                                   |
| provider/pkg/provider/provider_2e2_test.go                            | E2E test for the EFS replication scenario                             |
| provider/pkg/provider/testdata/efs-replication-protection/Pulumi.yaml | Test Pulumi program                                                   |

Fixes #2686

### DIFF
--- a/provider/pkg/provider/testdata/efs-replication-protection/Pulumi.yaml
+++ b/provider/pkg/provider/testdata/efs-replication-protection/Pulumi.yaml
@@ -31,6 +31,11 @@ resources:
         replicationOverwriteProtection: DISABLED
     options:
       provider: ${westNativeProvider}
+      # Ignore replicationConfiguration - AWS adds this to destination filesystems to show
+      # the replication relationship. Since we manage replication via the sidecar resource
+      # (efsReplication), we don't want Pulumi to try to reconcile this read-only property.
+      ignoreChanges:
+        - replicationConfiguration
 
   # Source filesystem - NO inline replicationConfiguration (use sidecar resource instead)
   efsSource:


rshade:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

---

**Review Summary:**
- Reviewed 5 files changed in this PR
- Checked for CLAUDE.md compliance against project rules
- Scanned for logic errors, nil pointer issues, and security vulnerabilities
- All potential issues identified were validated and determined to be either not applicable to this PR or not high signal

provider/pkg/provider/provider.go
```diff
@@ -1061,6 +1122,11 @@ func (p *cfnProvider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pu
 			diff := oldInputProjection.Diff(newInputProjection)
 			// 4. Apply this difference to the actual inputs (not a projection) that we have in state.
 			newInputs = resources.ApplyDiff(inputs, diff)
+			// 5. Filter out properties that are "effectively read-only" - properties with nested
+			// read-only paths that the user didn't specify. These shouldn't be treated as inputs
+			// because the user can't control the read-only nested values and attempting to remove
+			// them (when they appear after refresh but aren't in the user's program) can fail.
+			newInputs = schema.FilterEffectivelyReadOnlyProperties(newInputs, inputs, spec.ReadOnly)
```
corymhall: I don't think this is needed

corymhall (changes requested):
@rshade I think to fix the root cause here the _only_ thing we need is the diff suppression piece inside `Diff` right? I'm not sure I follow all the `read-only` stuff added to `Check` and `Read`. The property that causes this issue (`ReplicationOverwriteProtection `) is not a read-only property so it wouldn't even be effected by these changes  would it?

corymhall (changes requested):
I think we are almost there. I created a PR against this one to try and illustrate what I mean with some of my comments. You can view the diff here:

https://github.com/pulumi/pulumi-aws-native/pull/2706/changes/54f0dfd3a28a78738b483a70f4f168c4258d0518

PR #2796, opened 2026-02-02
feat: add enhanced hook error message handling

corymhall:


corymhall:
Superseded by PR #2827, which carries the rebased + review-addressed delta and test updates. Keeping #2796 as historical mirror context.

PR #2902, opened 2026-04-08
Don't nest the Region type

Frassle:
Core doesn't want to support modules nested under the "index" module. It makes little sense, and this appears to be the only instance of a type trying to do it.

This needs updating so we can tighten the token validation in the CLI to dis-allow patterns like this without breaking aws-native builds.


Frassle:
OK there doesn't seem to be a good path forward here. Changing the types to just be string means TS and Python change their API from enum to string. Fixing the refs to not be nested causes all the code generators to remove it from provider outputs (wat?!) and to change from strings to enums everywhere.

So either way this looks like a break change 😕

Frassle:
OK this is a bit grim but it keeps things working mostly as they were.

Turns out this type token was never bound correctly because it was in the type table as "aws-native:index/Region:Region" but referred to as "#/types/aws-native:index/region:Region". Note the change of case in "/region". As such it always hit the AllowDanglingRefs logic and bound as a `TokenType`.

Changing it to actually bind, and thus become an `EnumType` changes code generation substantially because of rules around provider outputs and non-string values. So we continue to abuse the case-sensitivity of type binding and declare this as "aws-native:index:Region" but refer to it as "aws-native:index:region". The schema now doesn't have any warnings about modules nested under index, and the only change to codegen is the python enumeration which writes out the new declared token ("aws-native:index:Region").

I think long term we'll want to change this to actually be a string enum, but we'll want to fix provider outputs to support that first.

corymhall (approved):

Issue #2944, opened 2026-04-24
CloudControl patch generation: send precise, updatable paths for AWS Native updates


corymhall:
Parent epic: #2942

## Summary

AWS Native update patch generation is currently too coarse for some nested properties. When a nested field changes, the provider can emit a patch for the top-level parent object, which may drag along unchanged nested siblings. If those siblings are create-only or otherwise not valid during update, CloudControl can reject the update.

This epic tracks making CloudControl patch generation more precise and metadata-aware.

## Problem

CloudControl accepts JSON Patch documents, but AWS Native currently often generates top-level patch operations from top-level input diffs. For complex inputs, that can produce broad replacements such as `/KubernetesNetworkConfig` instead of targeted paths such as `/KubernetesNetworkConfig/ElasticLoadBalancing/Enabled` or `/Version`.

This becomes especially risky when nested create-only/write-only properties are present.

## Related issues

Native sub-issues:

- #1722 - nested `writeOnly` / `createOnly` JSON patch generation
- #2936 - EKS cluster update failures involving nested create-only patch data
- #2260 - EKS nodegroup update patch shape

## Desired direction

Generate update patches from changed, updatable paths rather than broad parent objects where possible.

The implementation should account for:

- nested `createOnly` paths
- nested `writeOnly` paths
- unchanged sibling fields that should not be included in an update patch
- user-requested create-only changes that should be planned as replacement, not silently ignored
- resource handlers that may require unchanged contextual fields during update

## Success criteria

- Version-only updates do not send unrelated nested create-only properties.
- Nested mutable changes can be represented as nested patch operations where CloudControl supports them.
- Nested create-only/write-only metadata is respected during patch generation.
- Real changes to create-only properties are not silently dropped.
- Regression coverage exists for nested create-only/write-only patch cases.

Issue #2941, opened 2026-04-24
Import infers provider-observed/default fields as desired inputs


corymhall:
### What happened?

When importing AWS Native resources, the provider can infer CloudControl-observed/default fields as desired inputs. This can cause an immediate `pulumi preview --diff` after import to show updates or replacements even when the Pulumi program only specifies a narrower set of intended inputs.

The relevant path is the standard resource `Read` path for imports: when there are no prior `__inputs`, the provider calls `GetInputsFromState` on the CloudControl response. `GetInputsFromState` filters by top-level input property names only, so nested/default/output-shaped values returned by CloudControl can be stored in `__inputs` as if they were user-authored desired inputs.

### Example

Using an imported `aws-native:eks:Cluster`, the resulting `__inputs` included CloudControl-returned/default fields such as:

```json
{
  "accessConfig": { "authenticationMode": "CONFIG_MAP" },
  "controlPlaneScalingConfig": { "tier": "standard" },
  "deletionProtection": false,
  "encryptionConfig": [],
  "kubernetesNetworkConfig": {
    "elasticLoadBalancing": { "enabled": false },
    "ipFamily": "ipv4",
    "serviceIpv4Cidr": "10.1.0.0/16"
  },
  "logging": { "clusterLogging": { "enabledTypes": [] } },
  "resourcesVpcConfig": {
    "endpointPrivateAccess": false,
    "endpointPublicAccess": true,
    "publicAccessCidrs": ["0.0.0.0/0"],
    "securityGroupIds": [],
    "subnetIds": ["subnet-...", "subnet-..."]
  },
  "upgradePolicy": { "supportType": "EXTENDED" }
}
```

The Pulumi program only specified the intended inputs, for example `kubernetesNetworkConfig.serviceIpv4Cidr`, `resourcesVpcConfig.endpointPublicAccess/subnetIds`, `roleArn`, `tags`, and `version`.

After import, `pulumi preview --diff` showed updates/replacements caused by removing those imported/default fields from desired inputs. Similar default-field diffs appeared for related imported resources such as VPCs, subnets, routes, and IAM roles.

### Expected behavior

Import should avoid treating provider-observed/default CloudControl response fields as desired inputs. Outputs should continue to reflect CloudControl state, but `__inputs` should represent user-owned/imported desired inputs as accurately as possible.

### Related issues

This is related to, but distinct from:

- #2390: refresh after new input properties are added
- #884: refresh accuracy for complex properties
[...]

Issue #2940, opened 2026-04-24
[BUG] Pulumi refresh triggers adding `bearerTokenAuthenticationEnable` to every resource


Jimmy89:
### Describe what happened

Since the last few version (somewhere around 1.58.0 it started) till the current version: if I refresh my stack (`pulumi refresh -y -s stackName`) _every_ aws-native resource gets an additional `bearerTokenAuthenticationEnable: false` property added to both input as output.

If you accept these changes (which you need to because for example another resource got manually deleted) _and afterwards_ you up the stack, you get for _every_ resource an update to _delete_ this added property.
If you accept to run that change you get a response from the server back stating:

```
 aws-native:logs:LogGroup (Resource name):
    error: operation UPDATE failed with "ServiceInternalError": The security token included in the request is invalid (Service: CloudWatchLogs, Status Code: 400, Request ID: XXXXX) (SDK Attempt Count: 1)
```

It's really frustrating, as the manual fix is to export the current stack `pulumi stack export -s stackName > state.txt`, remove every single `bearerTokenAuthenticationEnabled: false` line, save and then re-import `pulumi stack import -s stackName --file state.txt`

After saving the update command works just fine, waiting for a problem at the next refresh.

This might be good to note: we use a AWS s3 bucket as hosting solution (self-hosted backend), we do not use Pulumi services.

### Sample program

```
import * as pulumi from "@pulumi/pulumi";
import * as awsCC from "@pulumi/aws-native";

const [
  { region },
  { accountId },
] = await Promise.all([
  awsCC.getRegion(),
  awsCC.getAccountId(),
]);

new awsCC.logs.LogGroup("loggroup-sns-tasks", {
  logGroupName: pulumi.concat("sns/", region, "/", accountId, "/name"),
  retentionInDays: 1
});
```

Up and refresh
Note: we use a AWS s3 bucket as hosting solution, we do not use Pulumi services.

### Log output

The `pulumi up` output:
```
 pulumi:pulumi:Stack                stackName                                          running..
 ~   ├─ aws-native:logs:LogGroup        ResourceName  update      [diff: -bearerTokenAuthenticationEnabled]
```

The pulumi details output:
```
   ~ aws-native:logs:LogGroup: (update)
        [id=/aws/vendedlogs/states/xxxxx]                                                                         
        [urn=urn:pulumi:xxx]
        [provider=urn:pulumi:XXX]
      - bearerTokenAuthenticationEnabled: false
```

[...]

PR #2956, opened 2026-05-01
Fix refresh input ownership semantics

### LINKED ISSUE #2390: Refresh doesn't handle new input properties
### Describe what happened

In `Read` we use the resource schema to map the newly read state to the inputs. When a new version of the provider is published that has _new_ optional input values that the user has not provided, the `Read` method reads the cloud, finds the new values and projects them onto the inputs. Then when update is run it has the value as an old input, but not in the current inputs (user program).

1. User deploys this program
```ts
new ccapi.rds.DbInstance('instance', {});
```
2. Provider is updated to add a new optional createOnly property called `backupTarget`
3. `pulumi refresh` is run
- `Read` finds a value for `backupTarget` in the cloud. Projects onto `oldInputs`
4. `pulumi up` is run
- Users program does not have `backupTarget`, but it is found in the prior inputs. Diff reported.

### Sample program

Deploy with `v1.29.0`

```ts
new ccapi.rds.DbInstance('instance', {});
```

Then upgrade to latest version and run `pulumi preview --refresh --run-program`

### Log output

_No response_

### Affected Resource(s)

_No response_

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

corymhall:
## Summary

Fixes #2390 by separating AWS Native refresh inputs from live output projection. Refresh now keeps managed drift visible to the engine while avoiding adoption of unowned optional input/output defaults into checkpointed `__inputs`.

This PR also updates the runtime paths that depend on that model:

- `Read` returns ownership-filtered refreshed inputs for existing standard resources, while outputs continue to reflect live CloudControl state.
- `Diff` and `Update` build their baseline from projected actual outputs where needed, with `__inputs` used as the ownership source.
- Nested path classification now handles required fields, read-only/write-only/create-only metadata, arrays, maps, and optional-computed-style input/output overlap.
- Write-only output rehydration is shared across `Create`, `Read`, and `Update`, including wildcard paths and omitted parent arrays.
- Create-only + write-only values are carried forward in output/checkpoint state, but are not forced into update patches.

## Scope notes

- This intentionally continues to use checkpointed `__inputs` as the standard-resource ownership source. Protocol `Inputs` / `OldInputs` adoption should happen in a follow-up PR.
- Import now uses recursive writable output projection instead of the old shallow `schema.GetInputsFromState` behavior. That excludes nested read-only and write-only values, but import can still infer provider-observed writable defaults; broader import ownership cleanup remains related to #2941.
- Precise nested CloudControl patch generation is not part of this PR. Nested write-only resend still flows through the existing top-level diff-to-patch path; metadata-aware leaf-level patch generation is tracked by #2944.

## Testing

```bash
cd provider
mise exec -- go test -count=1 ./pkg/resources -run 'TestPathClassifier|TestPathHelpers|TestExpandMatchingPaths|TestCalcPatch|TestSuppressAWSManagedDiffs|TestSuppressAWSManagedTagAdditions'
mise exec -- go test -count=1 ./pkg/provider -run 'TestStandardResourceDiffUsesActualOutputBaseline|TestStandardResourceReadReturnsOwnershipFilteredInputs|TestRead|TestUpdate'
```

Also ran:

```bash
git diff --check
```


corymhall:

- #2962
    - #2956 ◀

.agents/skills/aws-native-provider-runtime-design/agents/openai.yaml
guineveresaenger, line 1: Should the files in this folder be in a separate PR?

guineveresaenger (approved):
I think the overall approach makes sense.

What would migration look like for already-polluted state?

corymhall:
> What would migration look like for already-polluted state?

I decided not to handle that intentionally. I don't think there are many (if any) programs that currently have polluted state. One of the users that reported this issue in #2940 said they manually edited the state to get around the issue and I think that is what users that encountered this would have done. I don't think any users would run refresh, get polluted state, run preview and see a diff and just leave it.

PR #2962, opened 2026-05-05
Preserve Any property keys in CFN outputs

### LINKED ISSUE #2961: Make CloudFormation-to-SDK output conversion schema-aware for Any properties
## Summary

`naming.CfnToSdk` is schema-blind and recursively lower-camel-cases every map key returned by CloudControl. This is correct for normal AWS Native resource properties, but likely wrong for `pulumi.json#/Any` properties, where the nested object is domain JSON rather than AWS Native schema fields.

Write-side conversion already treats `pulumi.json#/Any` as opaque. Read/refresh conversion should probably do the same.

## Background

On writes, `naming.SdkToCfn` is schema-aware:

- top-level SDK property names are converted to CloudFormation names;
- typed object properties are converted recursively using schema;
- `Ref: "pulumi.json#/Any"` returns the value unchanged.

On reads/refreshes, the provider currently calls:

```go
rawState := naming.CfnToSdk(resourceState)
```

`CfnToSdk` does not receive schema information. It recursively walks all maps/arrays and applies `naming.ToSdkName` to every map key.

That means this CloudControl value:

```json
{
  "AssumeRolePolicyDocument": {
    "Version": "2012-10-17",
    "Statement": [{ "Action": "sts:AssumeRole" }]
  }
}
```

becomes this Pulumi state value:

```json
{
  "assumeRolePolicyDocument": {
    "version": "2012-10-17",
    "statement": [{ "action": "sts:AssumeRole" }]
  }
}
```

The outer conversion is correct. The inner IAM policy conversion is not: IAM policy keys are part of the policy JSON format, not AWS Native schema-property names.

## Why This Matters

The refresh/input ownership work in #2390 makes the provider use live CloudControl output more accurately as the old-input baseline for managed properties. That exposes representation differences that were previously hidden.

IAM Role policy documents are one observed manifestation, but this is probably broader. The schema contains many `pulumi.json#/Any` properties, including:

- IAM/S3/SNS/SQS/KMS/etc. policy documents
- API Gateway/OpenAPI bodies
- EventBridge event patterns
- CloudFormation templates
- SSM documents
- service-specific configuration blobs
- tags or parameter maps modeled as `Any`

For these values, recursively converting all nested keys to SDK casing can corrupt or normalize domain JSON in ways the provider cannot generally reason about.

## Proposed Fix

Introduce a schema-aware CloudFormation-to-SDK conversion path, roughly mirroring `SdkToCfn`.

Expected behavior:

- Convert top-level CloudFormation resource property names to SDK names.
- Recurse into typed object refs using `CloudAPIType` metadata.
- Recurse into arrays using item schema.
- Recurse into maps using `additionalProperties` schema.
- When the current type is `pulumi.json#/Any`, preserve the value unchanged.

Then use this schema-aware conversion in standard resource read/create/update output paths where resource schema is available.

## Compatibility Notes

This may cause a one-time state representation change for existing stacks whose `Any` outputs were previously lower-camel-cased by recursive `CfnToSdk`.

That one-time diff/state normalization is acceptable. After refresh writes the corrected shape into state, subsequent previews should be clean as long as the program input and AWS output are semantically aligned.

We should still add tests that make this behavior explicit.

## Suggested Tests

Add unit coverage for schema-aware `CfnToSdk`:

1. Normal typed nested resource properties still convert:
   - `RoleName -> roleName`
   - nested typed fields like `NetworkConfiguration.AwsvpcConfiguration.AssignPublicIp -> networkConfiguration.awsvpcConfiguration.assignPublicIp`

2. `pulumi.json#/Any` preserves nested keys:
   - `AssumeRolePolicyDocument.Version` stays `Version`
   - `Statement[].Action` stays `Action`

3. Arrays/maps around `Any` preserve `Any` element keys.

4. Representative resource tests for:
   - IAM Role `assumeRolePolicyDocument`
   - an inline `policyDocument`
   - an API Gateway/OpenAPI `body` or EventBridge `eventPattern`

## Open Questions

- Should invokes/function outputs also get a schema-aware conversion path where metadata is available?
- Are there any `Any` properties whose current lower-camel output shape is intentionally relied on? Write-side opaque handling suggests no, but this should be validated with tests.

corymhall:
## Summary
- add schema-aware CfnToSdkV2 conversion while preserving the existing CfnToSdk API
- use CfnToSdkV2 for standard Create, Read, and Update outputs
- preserve pulumi.json#/Any payload keys while still mapping known CFN resource/type properties to SDK names

## Why
The write path already uses schema metadata when converting SDK-shaped inputs to CloudFormation-shaped payloads. That lets it distinguish provider-typed objects from `pulumi.json#/Any` values, so user-defined JSON documents such as IAM policies keep keys like `Version`, `Statement`, and `Action` intact when sent to CloudControl.

The read/output path did not have the same schema-aware boundary. It used legacy `CfnToSdk`, which lower-camel-cased every nested object key. That is correct for typed provider objects, but wrong for arbitrary JSON payloads returned in `pulumi.json#/Any` properties, and can cause returned state to differ from the JSON users supplied.

This change makes output conversion symmetric with the existing write conversion while keeping the existing public `CfnToSdk` API unchanged. `CfnToSdkV2` lets provider runtime paths use schema metadata to map known CFN resource/type properties to SDK names while preserving arbitrary JSON payloads as-is.

## Testing
- mise exec -- go test -count=1 ./pkg/naming
- mise exec -- go test -count=1 ./pkg/provider -run 'TestRead|TestCreate|TestUpdate'
- mise exec -- make test_provider_fast

fixes https://github.com/pulumi/pulumi-aws-native/issues/2961

### DIFF
--- a/provider/pkg/naming/convert.go
+++ b/provider/pkg/naming/convert.go
@@ -11,11 +11,13 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/mattbaird/jsonpatch"
-	"github.com/pulumi/pulumi-aws-native/provider/pkg/metadata"
+
 	"github.com/pulumi/pulumi-go-provider/resourcex"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	"github.com/pulumi/pulumi-aws-native/provider/pkg/metadata"
 )
 
 // SdkToCfn converts Pulumi-SDK-shaped state to CloudFormation-shaped payload. In particular, SDK properties
@@ -43,6 +45,8 @@ var schemaStringReplacer = strings.NewReplacer(
 	// Paragraph separator
 	"\u2029", "\n")
 
+const schemaTypeObject = "object"
+
 // SanitizeCfnString ensures that a string from CFN docs meets the requirements for Pulumi schema strings.
 func SanitizeCfnString(str string) string {
 	return schemaStringReplacer.Replace(str)
@@ -51,6 +55,16 @@ func SanitizeCfnString(str string) string {
 type sdkToCfnConverter struct {
 	spec  *metadata.CloudAPIResource
 	types map[string]metadata.CloudAPIType
+}
+
+type cfnToSdkConverter struct {
+	spec  *metadata.CloudAPIResource
+	types map[string]metadata.CloudAPIType
+}
+
+type sdkProperty struct {
+	name string
+	spec pschema.PropertySpec
 }
 
 func (c *sdkToCfnConverter) sdkToCfn(properties map[string]interface{}) (map[string]interface{}, error) {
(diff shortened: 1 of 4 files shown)


corymhall:

- #2962 ◀
    - #2956

guineveresaenger (approved):

guineveresaenger:
Do we want to add this to Invoke() as well? doesn't have to be in this change

corymhall:
> Do we want to add this to Invoke() as well? doesn't have to be in this change

I did consider adding it to Invoke, but we currently don't have the plumbing in place to do the same sort of schema aware conversion. We can handle this in a follow up PR if Invokes run into issues related to this, but we don't have any reports of it yet.

PR #2978, opened 2026-05-13
Implement CloudControl List RPC semantics

### LINKED ISSUE #2976: Implement `List` for aws-native resources
[pulumi/pulumi#22693](https://github.com/pulumi/pulumi/pull/22693) added the [`List` RPC](https://github.com/pulumi/pulumi/blob/b8d0a0595a16729017728bc81a4cc0fd2e77d702/proto/pulumi/provider.proto#L175) (`pulumirpc.ResourceProvider.List`) and the [`listInputs` field on `ResourceSpec`](https://github.com/pulumi/pulumi/blob/b8d0a0595a16729017728bc81a4cc0fd2e77d702/pkg/codegen/schema/schema.go#L1877) to the provider protocol. `pulumi-aws-native` does not yet declare `listInputs` on its resources or serve `List`.

Work:

- [ ] Decide which underlying AWS API(s) to enumerate against, and which resource types to enable first.
- [ ] Define `listInputs` for the enabled types.
- [ ] Implement `List` dispatch, returning streamed `Result` and `Continuation` responses per the proto contract.
- [ ] Tests covering paginated and filter-driven enumeration.

Tracked by [pulumi/pulumi#3742](https://github.com/pulumi/pulumi/issues/3742). Engine support landed in [pulumi/pulumi#22693](https://github.com/pulumi/pulumi/pull/22693).

corymhall:
## Summary

Implements the AWS Native provider `List` RPC for CloudControl-backed resources, following the semantics documented in `docs/gh-2976-list-rpc-semantics.md`.

This PR adds schema-driven `listInputs`, CloudControl `ListResources` client support, provider-side query validation and SDK-to-CloudFormation casing conversion, and provider-owned continuation handling for paginated list calls.

Fixes #2976.

## What changed

- Generate `listInputs` for resources whose CloudFormation schema declares `handlers.list`.
- Preserve listability separately from list query shape with `HasListHandler`, since many listable resources have no `handlerSchema`.
- Add CloudControl `ListResources` wrappers in the client layer.
- Implement `cfnProvider.List` for standard CloudControl resources.
- Validate list queries against generated list input metadata before calling CloudControl.
- Convert Pulumi SDK-cased query inputs back to CloudFormation casing for `ResourceModel`.
- Stream CloudControl `ResourceDescription.Identifier` as the importable `ListResponse.Result.Id`.
- Add provider-owned continuation sessions so `page_size` controls CloudControl request chunking while `limit` is enforced across continuation requests.
- Regenerate provider schema and metadata to include `listInputs`.

## Design decisions

The detailed semantics are in `docs/gh-2976-list-rpc-semantics.md`, but the main decisions are:

- `List` is CloudControl-only. We do not add service-native AWS SDK list paths.
- `handlers.list` means a resource is listable; `handlers.list.handlerSchema` only describes optional query inputs.
- Resources with `handlers.list` but no `handlerSchema` get empty `listInputs`; non-empty queries for those resources are rejected.
- `ListResponse.Result.Id` is the CloudControl identifier because the protocol expects an importable ID.
- `ListResponse.Result.Name` is intentionally left empty for now because there is no generally correct AWS Native naming rule.
- The provider validates required query inputs and unknown fields itself instead of relying on engine-side validation.
- Unknown/computed query values return `ListResponse.Computed` without calling CloudControl.
- Continuation tokens are provider-owned rather than exposing CloudControl `NextToken` directly, so the provider can preserve request identity and enforce `limit` across pages.

## Review points

Please especially review:

- The `List` RPC behavior in `provider/pkg/provider/provider.go`.
- Continuation session lifecycle and validation in `provider/pkg/provider/list_sessions.go`.
- The generated `listInputs` model in `provider/pkg/schema/gen.go`.
- Query validation and SDK-to-CFN casing behavior.
- The decision to leave `Name` empty and return only importable IDs.
- The provider-owned continuation model, especially `page_size`, `limit`, expiration, and changed-request rejection semantics.

## Validation

- `mise exec -- make codegen generate_schema`
- `mise exec -- go test -count=1 ./pkg/provider -run 'TestList'`
- `mise exec -- go test -count=1 ./pkg/schema -run 'TestGatherList|TestList'`
- `mise exec -- golangci-lint run -c ../.golangci.yml --new-from-rev=origin/master ./pkg/metadata ./pkg/provider ./pkg/schema`
- `git diff --check`


provider/pkg/provider/list_sessions.go
```diff
@@ -0,0 +1,144 @@
...
+	"encoding/hex"
+	"sync"
+	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// defaultListSessionTTL is how long an unused provider List continuation token remains valid.
+const defaultListSessionTTL = 5 * time.Minute
```
guineveresaenger, line 14: Is there a way for a client to override?

provider/pkg/provider/provider.go
```diff
@@ -1238,6 +1243,296 @@ func (p *cfnProvider) Delete(ctx context.Context, req *pulumirpc.DeleteRequest)
...
+	if err != nil {
+		return nil, err
+	}
+	if len(failures) > 0 {
+		return nil, status.Error(codes.InvalidArgument, formatValidationFailures(failures))
+	}
+	if len(query) == 0 {
+		return nil, nil
+	}
+	return naming.SdkToCfn(listSpec, types, resourcex.Decode(query))
```
guineveresaenger: should this have a `status.Error` wrap?

provider/pkg/client/client.go
```diff
@@ -142,6 +155,37 @@ func (c *clientImpl) Read(ctx context.Context, typeName, identifier string) (res
...
+			return nil, nil, fmt.Errorf(
+				"listing resources of type %s: CloudControl returned a resource with an empty identifier", typeName)
+		}
+		identifiers = append(identifiers, *description.Identifier)
+	}
+	return identifiers, continuation, nil
+}
+
+// listMaxResultsForError formats an optional CloudControl MaxResults value for error messages.
+func listMaxResultsForError(maxResults *int32) string {
```
guineveresaenger: could we re-use `listMaxResultsForLog` somehow?

provider/pkg/provider/provider.go
```diff
@@ -1238,6 +1243,296 @@ func (p *cfnProvider) Delete(ctx context.Context, req *pulumirpc.DeleteRequest)
...
+		},
+	}); err != nil {
+		if req.GetContinuationToken() == "" {
+			p.ensureListSessions().remove(providerContinuationToken)
+		}
+		return err
+	}
+	session.bufferedIDs = bufferedIDs
+	session.returned = nextReturned
+	session.cloudControlNextToken = nextCloudControlToken
```
guineveresaenger, line 1397: could these be set before L# 1375 for both paths?

guineveresaenger (approved):
A question on the default timeout, and some nits, none blocking.

I think the additional design doc is good too.

provider/pkg/provider/list_sessions.go
corymhall: No, not today. The time picked is a little arbitrary, but I picked something that should be enough for a normal continuation

provider/pkg/provider/provider.go
corymhall: Updated!

provider/pkg/provider/provider.go
corymhall: When we have an existing token I think delaying the mutation until after the continuation response is sent is intentional. The session has already been acquired through `get(...)`, which marks it `inUse`, so another caller with the same continuation token cannot observe or mutate the session while this RPC is active.

The failure semantics are different between the two paths:

- For a new token, if sending the continuation response fails, we remove the newly-created token, so the client has no usable token to retry.
- For an existing token, the client already has the token it used for this request. If we advanced `bufferedIDs`, `returned`, and `cloudControlNextToken` before sending the continuation response, then a failed send followed by a retry with the same existing token could resume from the advanced state and skip results.

So the shape here is: initialize before exposure for new tokens, but for existing tokens only advance after the next continuation response is successfully sent.

provider/pkg/client/client.go
corymhall: Done. Added a shared helper

Issue #3066, opened 2026-07-07
Pulumi refresh sorts list alphabetically while pulumi up keeps the order of the program


raben2:
### Describe what happened

Pulumi refresh sorts lists for the sake of sorting them alphabetically, this causes resources to be updated during pulumi up even if there haven't been any changes. 

The example is based on iam role policies which can easily be prevented by naming them in a clear order, but on other resource types (e.g. vpc-endpoints) the sort order of attached elements (subnets) can not be hardcoded.

### Sample program

```
import * as aws_native from "@pulumi/aws-native";

const role = new aws_native.iam.Role("refresh-order-repro-role", {
  roleName: "refresh-order-repro-role",
  assumeRolePolicyDocument: {
    Version: "2012-10-17",
    Statement: [
      {
        Effect: "Allow",
        Principal: { Service: "lambda.amazonaws.com" },
        Action: "sts:AssumeRole",
      },
    ],
  },
  policies: [
    {
      policyName: "refresh-order-repro-policy",
      policyDocument: {
        Version: "2012-10-17",
        Statement: [
          {
            Sid: "",
            Effect: "Allow",
            Action: "logs:PutLogEvents",
            Resource: "*",
          },
          {
            Sid: "",
            Effect: "Allow",
            Action: "logs:CreateLogGroup",
            Resource: "*",
          },
          {
            Sid: "",
            Effect: "Allow",
            Action: "logs:CreateLogStream",
            Resource: "*",
          },
          {
            Sid: "",
            Effect: "Allow",
            Action: "logs:DeleteLogStream",
            Resource: "*",
          },
        ],
      },
    },
    {
      policyName: `fargateTaskExecutionSecretPolicy`,
      policyDocument: {
        Version: "2012-10-17",
        Statement: [
          {
            Sid: "01secrets",
            Action: [
              "secretsmanager:GetResourcePolicy",
              "secretsmanager:GetSecretValue",
              "secretsmanager:DescribeSecret",
              "secretsmanager:ListSecretVersionIds",
            ],
            Effect: "Allow",
            Resource: "*",
          },
        ],
      },
    },
    {
      policyName: `fargateTaskExecutionCloudwatchPolicy`,
      policyDocument: {
        Version: "2012-10-17",
        Statement: [
          {
            Action: ["logs:CreateLogStream", "logs:PutLogEvents"],
            Effect: "Allow",
            Sid: "01logs",
            Resource: "*",
          },
        ],
      },
    },
    {
[...]

PR #3082, opened 2026-07-17
feat(schema): preserve unordered collection metadata

corymhall:
## Why

CloudFormation knows whether an array is ordered or unordered through its
`insertionOrder` setting.

AWS Native currently throws that information away during schema generation. By
the time the provider compares a user's inputs with AWS state, it cannot tell
whether changing an array's order is meaningful.

This is the first half of the fix for #3066. It gives the runtime the information
needed by #3083 to avoid unnecessary updates when AWS returns an unordered array
in a different order.

## What this PR does

When a CloudFormation schema explicitly declares:

```json
"insertionOrder": false
```

the generator records that array's path in the resource's runtime metadata.

For example:

```text
policies
containerDefinitions/*/environment
```

The paths use Pulumi property names, `/` separators, and `*` for array elements.

The generated metadata currently records 2,850 unordered collection paths
across 1,416 resources.

## Why this is a separate PR

This PR only records metadata. It does not change refresh, Diff, updates, the
Pulumi package schema, or any generated SDK API.

Keeping metadata generation separate from runtime behavior makes it possible to
review whether we interpreted the CloudFormation schemas correctly before that
information affects users.

The runtime behavior is implemented in the stacked PR, #3083.

## Safety and tradeoffs

The generator only records arrays explicitly marked `insertionOrder: false`. It
does not guess based on property names, `uniqueItems`, or observed AWS behavior.

Schema references and ordinary alternatives are followed when the unordered
meaning is clear. Ambiguous schema combinations are skipped rather than
producing a path that might be wrong.

Generated paths are sorted and deduplicated so regeneration is deterministic.

Skipping an uncertain path means the provider may continue to report an
unnecessary order-only diff for that property. That is preferable to incorrectly
treating an ordered array as unordered and hiding a real change.

## Stack

This is the metadata foundation for #3083, which uses these paths during
refresh, Diff, and CloudControl patch generation.

Together, the two PRs address #3066.

## Validation

- `mise exec -- make codegen`
- `mise exec -- make generate_schema`
- Changed-lines `golangci-lint`: 0 issues
- Regeneration left the worktree clean

### DIFF
--- a/provider/pkg/metadata/metadata.go
+++ b/provider/pkg/metadata/metadata.go
@@ -41,7 +41,8 @@ type CloudAPIResource struct {
 	WriteOnly []string `json:"writeOnly,omitempty"`
 	// UnorderedCollections contains SDK-name slash paths for arrays whose
 	// CloudFormation schema declares insertionOrder: false. Array element
-	// traversal uses "*" path segments.
+	// traversal uses "*" path segments, including a terminal "*" when an array
+	// item is itself an unordered array.
 	UnorderedCollections []string               `json:"unorderedCollections,omitempty"`
 	IrreversibleNames    map[string]string      `json:"irreversibleNames,omitempty"`
 	TagsProperty         string                 `json:"tagsProperty,omitempty"`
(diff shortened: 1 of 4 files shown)


corymhall:

- #3082 ◀
    - #3083

docs/gh-3066-unordered-collection-semantics.md
```diff
@@ -0,0 +1,614 @@
+# GH-3066 Unordered Collection Runtime Semantics
```
corymhall: I'll remove this before merge.

pose (approved):

PR #3083, opened 2026-07-17
fix(resources): honor unordered collection semantics

### LINKED ISSUE #3066: Pulumi refresh sorts list alphabetically while pulumi up keeps the order of the program

corymhall:
## Why

CloudFormation marks some arrays as unordered. AWS may return those arrays in a
different order than the user supplied, even though their contents have not
changed.

AWS Native currently treats every array as ordered. During refresh, it can save
AWS's ordering as the new input state. The next preview then proposes an update
to restore the user's ordering.

For example:

```text
Program: [z-policy, a-policy]
AWS:     [a-policy, z-policy]
```

These policies mean the same thing, but the provider currently sees a change.
This causes repeated, unnecessary updates such as #3066.

## What this PR does

This PR uses the unordered-array metadata generated by #3082 when comparing
desired inputs with AWS state.

When an unordered array only changed order:

- refresh keeps the user's ordering in the saved inputs;
- outputs still show the order returned by AWS;
- Diff reports no change; and
- no CloudControl patch is generated.

The same comparison is used by refresh, Diff, and patch generation so these
paths cannot disagree.

## Safety and tradeoffs

The provider suppresses a difference only when it can pair every old element
with exactly one equivalent new element.

This means:

- duplicate values are counted rather than collapsed;
- additions, removals, and value changes remain real changes;
- ordered arrays are unaffected; and
- IAM policy documents can use IAM-specific semantic equality while matching.

When the provider cannot prove equality, it keeps the existing diff. This
includes values containing unknowns, element-level secrets, nested write-only
properties, and values that would require a CloudFormation property transform
to compare.

That conservative fallback may leave a phantom diff in some uncommon cases, but
it is safer than hiding real drift or associating a secret/write-only value with
the wrong array element.

This PR also does not attempt to produce minimal element-level patches for
changed unordered arrays. A real change may still replace the whole collection.

## Stack

Depends on #3082, which records CloudFormation's unordered-array metadata
without changing runtime behavior.

Fixes #3066.

## Validation

- Changed-lines `golangci-lint`: 0 issues
- Provider lifecycle tests for refresh and Diff
- Patch-generation tests
- Credentialed IAM Role and ECS TaskDefinition refresh test


corymhall:

- #3082
    - #3083 ◀

examples/unordered-collections/index.ts
```diff
@@ -0,0 +1,61 @@
+import * as pulumi from "@pulumi/pulumi";
```
pose: @pulumi-eon could we make this into a YAML-only test instead of having to have a TypeScript/JavaScript/NodeJS package? cc: @corymhall

pose (approved):

PR #3102, opened 2026-07-24
fix(schema): generate referenced object union properties

### LINKED ISSUE #3099: `aws-native.devopsagent.Association` resource is unusable b/c of `anyOf` in the upstream schema
The `configuration` property on `aws-native.devopsagent.Association` is, per upstream docs, a union type. We don't generate any supporting types docs for it, so it's hard to know what the proper usage is.

Registry page: https://www.pulumi.com/registry/packages/aws-native/api-docs/devopsagent/association/#configuration_nodejs
Upstream docs: https://docs.aws.amazon.com/devopsagent/latest/APIReference/API_ServiceConfiguration.html

corymhall:
## Summary

Referenced CloudFormation definitions that explicitly declare `type: object` are currently handled as ordinary objects before their `oneOf` or `anyOf` branches are examined. When all properties live in those branches, this generates an empty Pulumi type. For example, `devopsagent.Association.configuration` rejects `{ aws: ... }` as an unknown property even though it is a valid upstream branch.

This change merges non-conflicting branch properties into one named object:

```text
{ aws: AwsConfiguration } | { gitHub: GitHubConfiguration }

becomes

{ aws?: AwsConfiguration, gitHub?: GitHubConfiguration }
```

Unconditional requirements from the outer object remain required. Branch-specific requirements become optional because they are conditional on the selected branch; AWS Cloud Control remains responsible for enforcing the `oneOf`/`anyOf` constraint. If branches define the same property incompatibly, codegen conservatively falls back to the existing real-union representation.

Fixes #3099.

## Scope

The behavior applies generally to referenced, explicitly typed object unions rather than adding a DevOpsAgent-specific overlay. Regeneration updates the 22 existing types that can be safely merged and emits their previously missing nested SDK types. Existing untyped unions retain their prior generation path. The newly reachable SageMaker instance-storage FSx type uses a distinct name so it does not collide with the existing `ClusterFSxLustreConfig` when SDKs normalize acronym casing.

Provider runtime behavior and local validation of exactly-one/at-least-one constraints are intentionally unchanged.

## Validation

- [x] `mise exec -- make lint`
- [x] `mise exec -- make test_provider_fast`
- [ ] `mise exec -- make test_provider` (not run; the full suite may require live AWS credentials)
- [ ] `mise exec -- make test` (not run; integration contracts are unchanged)
- [x] `cd provider && mise exec -- go test -count=1 ./pkg/schema`
- [x] `mise exec -- make build_nodejs build_python build_dotnet build_java build_go`

## Generation / Drift Checks

- [x] No hand-edits to generated files, including `sdk/**` and generated schema artifacts
- [x] Ran `mise exec -- make generate_schema`
- [x] Ran `mise exec -- make local_generate`
- [ ] Ran `make ci-mgmt` (not applicable; `.ci-mgmt.yaml` is unchanged)

## Risk

The generated SDK surface expands for object unions that previously appeared empty. The merged type intentionally permits zero or multiple branch properties locally; Cloud Control will reject combinations that violate the upstream union. Property-name conflicts do not use this representation and preserve the existing union fallback.

## Rollback

Revert commits `890c3daa18` and `a7254e4bee` in that order; they include the corresponding generated schema, metadata, and SDK outputs.


pose (approved):
I am surprised that we need this to do "by hand" and engine is not able to solve this on its own.

PR #3113, opened 2026-08-01
Fix secret nesting amplification for write-only properties on refresh

yukimotochern:
## What

Refreshing a resource with a secret value under a write-only path (e.g. `AWS::Synthetics::Canary` `runConfig/environmentVariables`) doubles the value's secret wrapper depth on every refresh: 1 → 2 → 4 → 8 → 16 → … Each wrapper roughly doubles the JSON-escaped payload in the encrypted state, so a 36-character secret grew to ~438 KiB per occurrence after four refreshes on a production stack, and eventually `pulumi refresh` was OOM-killed.

Two defects combine:

1. `preserveSecretWrapper` re-wraps values that are already secret. Write-only values restored from old inputs by `AddWriteOnlyOutputFallbacks` are already secret, so each refresh stacks the old wrapper depth on top.
2. `addWriteOnlyFallbacks` aliases the old-input subtree into the new output state instead of copying, so the re-wrap also writes through into the parsed old inputs and flows into the new `__inputs` checkpoint — compounding across refreshes.

## Fix

- `preserveSecretWrapper` never wraps an already-secret value.
- `addWriteOnlyFallbacks` clones restored values instead of aliasing old inputs.
- `CollapseConsecutiveSecrets` collapses `Secret(Secret(x))` → `Secret(x)` at the checkpoint read/write boundary (`ParseCheckpointObject` / `CheckpointPropertyMap`), so states already poisoned by earlier versions heal on their next refresh.

## Tests

- Refresh-loop regression test asserting wrapper depth stays exactly 1 in outputs and `__inputs` over 20 simulated refreshes (fails with depth 2 on master at iteration 0).
- Healing test: a state with a 16-deep nested secret collapses to depth 1 after one refresh.
- Unit tests for the collapse at the checkpoint boundary, the no-nest behavior of `PreserveSecretWrappers`, and the no-alias behavior of `AddWriteOnlyOutputFallbacks`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/provider/pkg/resources/checkpoint.go
+++ b/provider/pkg/resources/checkpoint.go
@@ -12,23 +12,61 @@ func CheckpointObject(inputs resource.PropertyMap, outputs map[string]interface{
 }
 
 // CheckpointPropertyMap puts inputs in the `__inputs` field of the state.
+// Consecutive secret wrappers are collapsed so that checkpointed state never
+// grows an extra layer of secret nesting per operation.
 func CheckpointPropertyMap(inputs resource.PropertyMap, outputs resource.PropertyMap) resource.PropertyMap {
-	var props resource.PropertyMap
-	if outputs == nil {
-		props = resource.PropertyMap{}
-	} else {
-		props = outputs.Copy()
+	props := resource.PropertyMap{}
+	for key, value := range outputs {
+		props[key] = CollapseConsecutiveSecrets(value)
 	}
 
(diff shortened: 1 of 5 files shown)


corymhall:
@yukimotochern thanks for the PR! I modified your solutions slightly and merged it in https://github.com/pulumi/pulumi-aws-native/pull/3115. I also kicked of a release so a new version should be out soon.

